PR #1, opened 2026-07-07
DSP, Virtual DSP / Time Alignment and audio-capture fixes from full code review

DIMOSUS:
Results of a full review of the DSP library, the Virtual DSP / Time Alignment tools and the audio capture layer, with fixes, deduplication, performance work and new tests. Also adds `CLAUDE.md` (build commands + architecture overview for AI-assisted sessions).

## DSP library (`dsp/`)

**Bug fixes**
- `CalibrationFile`: clamped interpolation below the first calibration point — unbounded extrapolation could drive corrections to the −160 dB floor (a +160 dB spike on the plot); nearest-point fallback for degenerate Lanczos weight sums; new `HasData` so a missing/unparsable file is detectable.
- `LogarithmicResample`/`SmoothLinear`: degenerate kernel windows now hold the nearest sample instead of collapsing to the dB floor (e.g. resampling to 20 kHz from a spectrum that ends below it).
- `Windowing.TukeyWindow`: normalizes negative/overlapping fades, explicit zero-fade handling instead of the `!IsFinite → 1` patch.
- `EasyEffectsFormat`: removed the bogus `width = Q` field (verified against real EasyEffects presets — Bell bands carry no `width`).

**Performance**
- `GetAutocorrelation`: Wiener–Khinchin FFT autocorrelation in double precision replaces the O(n²)-per-lag float loop (~100× fewer operations per redraw).
- Virtual DSP alignment searches: coarse delay grids evaluated with incremental phasor rotation instead of `Complex.Exp` per (bin × delay).

**Deduplication**: single Lanczos kernel in `DspMath` (was ×3), shared `CurveSampling.InterpolateDbLog` (two identical 40-line copies), `GetPhaseData` via `BuildMeasuredPhase`, shared H1 core for both `ComputeRelativeIr` variants, `PeqBand.IsTransparent`, and more.

## Virtual DSP / Time Alignment (`source/`)

- One speed-of-sound constant (343.2 m/s at 20 °C) — there were four declarations with two different values across the tools.
- Auto delay: junction searches use a thread-confined FFT cache (channels with unchanged chains are no longer re-FFT'd on every search step); failures surface an error dialog instead of a beep.
- `Clipboard.SetText` guarded (busy clipboard crashed the app); manually created `ToolTip`s disposed; envelope preview uses min/max decimation so narrow reflection peaks stay visible; PDF sheet gets 8 print-friendly channel colours matching the on-screen hues.
- Testability extraction: `AlignmentSelection` (dsp), `VirtualCrossoverJunctions`, `VirtualCrossoverMetric`, `DelayTableText` — the fragile pure logic now lives in unit-tested classes.

## Audio capture layer — ASIO stutter fixes (`source/Audio/`, `source/Measurements/`)

NAudio fills ASIO output buffers only after `AudioAvailable` returns, so everything slow in the callback delayed playback:
- **`CaptureAccumulator`** — shared accumulation core with pre-allocated capacity (sweep length is known up front) and consumed-tail trimming in live mode: bounded constant memory instead of unbounded `List&lt;float&gt;` growth with in-callback re-allocations and LOH/GC pauses.
- **`AsioSampleConverter`** — bulk conversion (one `Marshal.Copy` + one typed pass) on reusable scratch replaces per-sample `Marshal.ReadInt32`; conversion and level metering moved outside the lock.
- `LoopingWaveProvider` no longer loops forever on an empty source; Wave live noise plays seamlessly through the looping provider (the old play-to-end loop left an audible gap every pass).
- `ConfigureAwait(false)` across the ASIO path (closing the app mid-measurement could deadlock the blocking `Dispose`); live-capture cancellation reported as success; `CaptureChannelLayout` deduplicates channel-window arithmetic copied across three classes.

## Tests

- DSP: 263 passing locally (was 237), including new coverage for calibration extrapolation, Tukey windows, FFT autocorrelation equivalence (1e-9 vs direct computation) and alignment-candidate selection.
- App: ~30 new tests (metric formatting, junction arithmetic, delay-table parsing, capture accumulator, ASIO sample conversion via pinned buffers, looping provider). **These run on Windows CI only** — this PR is their first execution.
- Full solution and App.Tests compile with 0 errors / 0 warnings (cross-compiled with `EnableWindowsTargeting`).

## Worth a manual check on Windows

- A sweep + a longer Live Spectrum session over ASIO: the stutter should be gone/much reduced, memory flat.
- Seamless Wave noise loop; tuning-sheet mm read-outs (343.0 → 343.2, e.g. 0.42 ms still prints 144.1 mm).
- Crossover wizard layout at 150 %+ DPI (pre-existing concern, untouched here).

https://claude.ai/code/session_01KP55go5gqByy6nRGAVAjGv

---
_Generated by [Claude Code](https://claude.ai/code/session_01KP55go5gqByy6nRGAVAjGv)_


DIMOSUS:
Review note:

[P1] В `ReceiveAudio` всё ещё создаются новые `double[]` для `peaks` и `sumSquares` на каждый ASIO buffer-switch callback. Этот путь выполняется в real-time аудио callback, а PR как раз пытается убрать GC pressure/stutter из callback. При малых ASIO buffer sizes такие per-callback аллокации могут снова вызывать Gen0 collections и возвращать слышимые stutter'ы.

Предлагаю сделать эти массивы переиспользуемыми scratch buffers на уровне `AsioFullDuplexSession` (как `convertScratch`), ресайзить только при изменении числа каналов и очищать через `Array.Clear` перед измерением уровней.

DIMOSUS:
Fixed in 7686f55 — `peaks`/`sumSquares` are now instance scratch arrays in both `AsioFullDuplexSession.ReceiveAudio` and `SoundRecorder.ReceiveData`, resized only when the channel count changes (`SoundRecorder` additionally `Array.Clear`s them, since its loop accumulates instead of assigning). Reuse is safe: `MeasureChannels` copies the values into a fresh `AudioChannelLevel[]` before the next callback can overwrite the scratch, and callbacks are serialized on the driver thread.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2, opened 2026-07-07
Full review, stages 1–6: races, data safety, update hardening, controls & plotting fixes

DIMOSUS:
All six stages of the remaining-areas code review, one commit per stage.

## Stage 1 — Shell / mode switching / UI marshaling
- **Mode switches serialized** (`ModeController.SelectAsync` re-entrancy could leave `ActiveTab`/`CurrentMode` describing different modes); a failed switch doesn't break the chain.
- **Double-click on Record no longer crashes** (`Init` on a running measurement in an `async void` handler).
- **Level metering no longer floods the UI queue** — `CoalescingDispatcher<T>` replaces per-audio-callback `BeginInvoke`.
- Guard→BeginInvoke race closed (`TryBeginInvokeOnUiThread`); history activation during a sweep ignored; `FormClosing` re-entry guarded.

## Stage 2 — History / impulse-response files
- **Atomic saves** for `ImpulseResponseFile` and the history store (truncate-then-write destroyed data on a crash mid-write).
- **Bounded snapshot memory** — at most one file-backed entry keeps its full IR snapshot cached; injectable store path for tests.

## Stage 3 — Auto-update path / settings
- **No unsigned-update fallback** (always `SecurityMode.Strict`; keyless builds go to the release page; CI fails on an empty `SPARKLE_PUBLIC_KEY`).
- **Release URL validated** before ShellExecute (https + github.com only); `MeasurementSettingsFile.Save` atomic.

## Stage 4 — Signal sources / Signal Generator
- **`PcmStreamSet`**: three identical PCM-packing copies deduplicated; per-routing streams built lazily (~60 MB saved on long noise buffers); no per-sample allocation.
- **10 ms raised-cosine ramps** on generator output (click-free, tweeter-friendly edges); `PlaybackStopped` marshal guarded.

## Stage 5 — EQ Wizard
- **Auto Tune runs on a worker thread** (32-band fits froze the UI); failures show an error instead of crashing the handler.
- NaN-safe control writes via `ClampValue` (imported-profile NaN previously threw `OverflowException`).

## Stage 6 — Custom controls / Plotting / Overlays
- **`NumericTextParser`**: a lone `.`/`,` is the decimal separator in any culture — `decimal.TryParse` group-separator leniency parsed `1.5` in a comma-decimal locale as **15** (silently 10× wrong, then clamped). Culture thousands (`12,000` en-US) still round-trip.
- **Up/Down/wheel commit typed text before stepping** (previously overwrote it with lastValue ± step); **Enter commits before a dialog's AcceptButton** reads the value (it read the stale one).
- **Waterfall**: tracker args were shifted one position (X blank, Y = frequency, "Iterations" = dB); empty burst-decay slices no longer throw inside `Render`; slices aren't re-resampled on every pan/resize (view-window + revision cache); labels invariant-culture.
- **Live-spectrum coherence gating matched by frequency, not index** — the magnitude and coherence resample grids differ (bounds + skipped empty bands), so trusted/dimmed segments landed on the wrong frequencies; degenerate coherence can no longer blank the trace.
- **Overlays**: offset spinner no longer serializes the slot + fsyncs per tick (500 ms debounce, flushed before mode-switch reload); only captured-slot changes rebuild calculated overlays (a target offset re-ran ComplexSum analysis per tick); text export writes temp + move.

## Tests
Stage 1: ModeController (4), CoalescingDispatcher (5). Stage 2: atomic IR save (2), history persistence (4), snapshot cache (2). Stage 3: trusted-URL predicate (12 cases). Stage 4: PCM packing (5), fades (3). Stage 6: NumericTextParser (en-US/de-DE/ru-RU semantics).
Solution + App.Tests compile 0/0 (`EnableWindowsTargeting`); 263 DSP tests pass locally; app tests run on Windows CI here.

## Reviewed, reported, intentionally not changed
- Shell: `UpdatePeakInfo` unsynchronized pair read; Options dialogs' `FormClosed` unsubscribe misses handle-less closes; `WireLiveApply` wires only dialog-open controls; `CloseReason.WindowsShutDown` ignored.
- Update path: version comparison ignores prerelease identifiers (`-rc.1` → `-rc.2` doesn't prompt); uninstaller leaves settings behind.
- Controls: per-paint Pen/Brush churn in dark controls; `DarkComboBox` swallows Enter (no AcceptButton path from a combo) and doesn't take focus on click; `BeginInit` is a no-op (property-order-sensitive clamping); tools menu rebuilt per open.
- Plotting: live-path allocation churn (fresh series + several list copies per 30 fps tick); `fftLength` reconstruction from coherence length off by 2 (systematic tiny frequency skew); `LogarithmicClipAxis` label trim; dead `FourierCSD` enum value.
- Overlays: 30 fps target render rebuilds constant target/tolerance curves each tick (caching + extracting pure render math from `Overlay.cs` is the highest-value future refactor, with `OverlaySlotState` to kill the triple field-mapping); 12 JSON files re-read on every mode switch; amplitude-space A−B clamps negative results to −160 dB instead of a gap; corrupt slot files silently present as empty and can be overwritten.

https://claude.ai/code/session_01KP55go5gqByy6nRGAVAjGv

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,6 +94,10 @@ jobs:
         env:
           SPARKLE_PUBLIC_KEY: ${{ secrets.SPARKLE_PUBLIC_KEY }}
         run: |
+          if ([string]::IsNullOrWhiteSpace($env:SPARKLE_PUBLIC_KEY)) {
+            throw "SPARKLE_PUBLIC_KEY repository secret is required: a build without the embedded key cannot verify update signatures."
+          }
+
           $version = "${{ steps.version.outputs.version }}"
           $publishDirectory = Join-Path $env:RUNNER_TEMP "publish-${{ matrix.rid }}"
 
@@ -152,6 +156,10 @@ jobs:
         env:
           SPARKLE_PUBLIC_KEY: ${{ secrets.SPARKLE_PUBLIC_KEY }}
         run: |
+          if ([string]::IsNullOrWhiteSpace($env:SPARKLE_PUBLIC_KEY)) {
+            throw "SPARKLE_PUBLIC_KEY repository secret is required: a build without the embedded key cannot verify update signatures."
+          }
+
           $version = "${{ needs.prepare.outputs.version }}"
           $publishDirectory = Join-Path $env:RUNNER_TEMP "publish-win-x64"
 
--- a/source/Audio/PcmStreamSet.cs
+++ b/source/Audio/PcmStreamSet.cs
@@ -1,0 +1,141 @@
+using NAudio.Wave;
+
+namespace Resonalyze;
+
+/// <summary>
+/// Packs a mono float signal into PCM playback streams, one per
+/// <see cref="PlaybackChannel"/> routing, built lazily on first request: a
+/// measurement uses exactly one routing, while the previous eager four-way
+/// packing kept four PCM copies of the signal (tens of megabytes for long
+/// sweeps and noise buffers) alive for its whole lifetime.
+/// </summary>
+internal sealed class PcmStreamSet : IDisposable
+{
+    private static readonly int ChannelModeCount =
+        Enum.GetValues<PlaybackChannel>().Length;
+
+    private readonly float[] samples;
+    private readonly int sampleRate;
+    private readonly int bitsPerSample;
+    private readonly MemoryStream?[] memoryStreams = new MemoryStream?[ChannelModeCount];
+    private readonly RawSourceWaveStream?[] sourceStreams =
+        new RawSourceWaveStream?[ChannelModeCount];
+    private bool disposed;
+
+    public PcmStreamSet(float[] samples, int sampleRate, int bitsPerSample)
+    {
+        ArgumentNullException.ThrowIfNull(samples);
+        if (sampleRate <= 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(sampleRate));
+        }
+        if (bitsPerSample is not (16 or 24))
+        {
+            throw new NotSupportedException(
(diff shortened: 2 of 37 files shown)


DIMOSUS:
Нашёл потенциальный регресс в debounce-сохранении overlay offset.

В `Overlay.OffsetControlChanged` сохранение offset теперь откладывается через `offsetSaveTimer` на 500 мс. Flush происходит только через `OffsetSaveTimerTick` / `FlushPendingOffsetSave`. При этом в close path (`Form1_FormClosing`) вызывается `PersistCurrentSessionState()`, но pending overlay offset-save нигде явно не flush’ится.

Сценарий:
1. Пользователь меняет overlay offset через spinner.
2. Закрывает приложение быстрее чем через 500 мс.
3. Таймер не успевает сработать.
4. Последнее изменение offset не попадает в slot-файл.

Это выглядит как регресс относительно прежнего синхронного `TrySaveCurrentState(...)` при каждом изменении offset.

Предлагаю добавить что-то вроде `OverlayCollection.FlushPendingSaves()` и вызвать его в close path перед teardown / `PersistCurrentSessionState()`, либо flush’ить pending overlay saves при закрытии/disposing overlays.

DIMOSUS:
Fixed in 67ddc21 — `OverlayCollection.FlushPendingSaves()` now runs in `Form1_FormClosing` before `PersistCurrentSessionState()`, so an offset changed inside the 500 ms debounce window persists on close. Mode switches were already covered (each `Overlay.Prepare` flushes before reloading the slot from disk); the close path was indeed the missing one.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #59, opened 2026-07-25
EQ Wizard: calibration and smoothing for dB SPL captures

DIMOSUS:
A dB SPL capture — the Live Spectrum RTA the moving-microphone workflow produces —
stores its drawn curve rather than a raw spectrum, because its band levels cannot be
re-gridded to the display range without inventing bands the analyzer never resolved.
That left the calibration **and** the smoothing selector greyed out in the EQ Wizard for
exactly those curves.

Neither has to be a dead end.

## Calibration

These modes apply the microphone correction **additively per frequency** (the SPL RTA
subtracts it per band, then adds a constant SPL offset that comes from a *separate* SPL
calibration and does not depend on the 0°/90° choice). So the correction frozen at
capture can be taken back out exactly and another applied in its place — no raw spectrum
needed. `Own` reproduces the capture losslessly.

## Smoothing

A curve captured with smoothing Off **is** an unsmoothed reference, so it may be smoothed
here. One captured under smoothing — or one that never said (a text import, a legacy
slot) — is left alone, because smoothing an already smoothed curve compounds it.

## How

Two additive overlay fields, no file version bump:

- `PointsCalibrationCorrectionDb` — frozen per **drawn point**, since those frequencies
  are the only grid such a capture has; validated against the point count.
- `CapturedSmoothingCode` — null in legacy files, meaning "unknown", so nothing is assumed.

`EqWizardImportedCurve` renders the curve the way the measuring mode would: uncalibrate →
smooth **on its own frequencies** → calibrate with the chosen profile. No resampling, so
the display range is never extrapolated; NaN gaps survive every step.

One trap worth naming: the correction is aligned by index, and point normalization sorts
and drops duplicates. Normalizing the two separately would silently shift the calibration
in frequency, so they are normalized together.

## Also in here

- **Restores the EQ Filters selector**, which the sample-rate row added in #55 had placed
  at identical coordinates and covered in the z-order. The control was never removed —
  just hidden underneath.
- **A Reset filters button**: every band back to its default frequency at Q 1 and 0 dB,
  preamp 0, one active filter. The source, target and Auto Tune settings are kept — it
  clears the tune, not the setup. There is no undo here, so it asks first.

## Notes

Existing overlay files predate the annotations and stay as they are; re-capture a slot to
get the selectors.

## Testing

`EqWizardImportedCurveTests` (exact profile swap, lossless `Own` round trip, smoothing on
the curve's own grid, NaN gaps, mismatched correction ignored), resolver gating and paired
normalization, overlay round trip and validation. Panel layout re-checked for control
overlaps. DSP 875, App 682 green; builds with no warnings.

### DIFF
--- a/dsp/DataHelper.Resampling.cs
+++ b/dsp/DataHelper.Resampling.cs
@@ -453,39 +453,142 @@ namespace Resonalyze.Dsp
                 return output;
             }
 
-            var prefix = new double[steps + 1];
+            double[] smoothedAmplitudes = SmoothBandPowersToAmplitudes(
+                bandPowers, frequencies, octavesPerStep, smoothingHalfSteps, psychoacoustic);
             for (int i = 0; i < steps; i++)
             {
-                prefix[i + 1] = prefix[i] + bandPowers[i];
+                output.Add(new SignalPoint(
+                    frequencies[i],
+                    AmplitudeToDecibels(smoothedAmplitudes[i])));
+            }
+
+            return output;
+        }
+
+        /// <summary>
+        /// Re-applies this analyzer's display smoothing to band levels it already produced
+        /// — the finished dB curve of a dB SPL RTA, on the very grid it was drawn on.
+        /// </summary>
+        /// <remarks>
+        /// A consumer that stores such a curve (an overlay slot, and through it the EQ
+        /// Wizard) has no raw spectrum to go back to, but the smoothing is a SECOND PASS
+        /// over the band powers, not part of the integration — so it can be replayed
+        /// exactly. It shares its core with <see cref="LogarithmicPowerBandResample"/>,
+        /// which is the point: a level-preserving mean of linear POWER, not of decibels,
+        /// with the same window and the same psychoacoustic cubic mean, so "1/6 octave"
+        /// means the same thing in both places and cannot drift apart later.
+        /// A non-finite level marks a band the analyzer could not measure: it is passed
+        /// through and excluded from its neighbours' means, so gaps neither spread nor fill.
+        /// Points must be ascending and logarithmically spaced, as that grid is.
+        /// </remarks>
+        public static List<SignalPoint> SmoothBandLevels(
+            IReadOnlyList<SignalPoint> bandLevelsDb,
+            double smoothingOctaves,
+            bool psychoacoustic)
+        {
+            ArgumentNullException.ThrowIfNull(bandLevelsDb);
+
+            int steps = bandLevelsDb.Count;
+            var result = new List<SignalPoint>(steps);
+            if (steps < 2 || bandLevelsDb[0].X <= 0 || bandLevelsDb[^1].X <= bandLevelsDb[0].X)
+            {
+                result.AddRange(bandLevelsDb);
+                return result;
+            }
+
+            double octavesPerStep =
+                Math.Log2(bandLevelsDb[^1].X / bandLevelsDb[0].X) / (steps - 1);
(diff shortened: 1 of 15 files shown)


DIMOSUS:
Confirmed and fixed in b8c8abb — thanks, this was a real defect.

Verified against the sources: `LogarithmicResample` unpacks dB to linear amplitude
before averaging (`dBUnpack: true`), and `LogarithmicPowerBandResample` integrates band
power, while `OverlayMath.SmoothByOctaves` averages the dB values themselves for its
plain fractional-octave widths — a geometric mean, which flattens narrow peaks harder
and fills dips more than the mode that drew the curve. Auto Tune was being fed a shape
the measurement never had at that width.

One nuance worth recording: the psychoacoustic width was never affected. It already
takes a cubic mean of linear amplitudes, so only the plain 1/N widths were wrong.

The fix makes the domain explicit rather than picking a compromise:

- `SmoothByOctaves` gains a `MagnitudeAveraging` parameter (`Decibel` / `Amplitude` /
  `Power`), defaulting to `Decibel` so every existing overlay renders exactly as before.
- The EQ import path states the domain of the mode the curve came from, selected by its
  curve kind: `InputSpectrum` → `Power` (the RTA's band levels are power integrals),
  everything else → `Amplitude` (a swept response). So it reproduces that analyzer rather
  than a single hand-picked linear domain.

The existing ripple test now pins the property this turns on: a symmetric ±5 dB ripple
must settle *above* its arithmetic midpoint, because the loud half carries far more
energy — a decibel mean lands exactly on it, so the test discriminates the two.

App 683, DSP 875 green.

DIMOSUS:
Both points are now addressed — `b8c8abb` (domain) and `59e7b72` (algorithm). The
review landed on `46d7e04`, so it predates the first of those, but the deeper point stood
and was the one that mattered: **the domain is not the algorithm.**

Reading `LogarithmicPowerBandResample` gave the exact gap:

| | analyzer (RTA) | wizard, before |
|---|---|---|
| plain 1/N | uniform moving mean of band **powers** | raised-cosine weighted mean |
| psychoacoustic | cubic mean of **powers** | cubic mean of **amplitudes** |

The useful discovery is the one you pointed at: that smoothing is a **second pass over
the finished band powers**, not part of the integration — so it can be replayed exactly
on a stored curve, no raw spectrum needed.

So `DataHelper.SmoothBandLevels` now shares its core with the resampler: they are the
same code, and "1/6 octave" cannot come to mean two things. Prefix sums carry a count, so
an unmeasured band contributes nothing rather than poisoning a mean; with no gaps the
resampler's output is bit-identical to before.

The equivalence is asserted directly rather than by proxy: resample at a width, versus
resample unsmoothed then replay that width — equal to **1e-9 dB** across four widths
including the psychoacoustic one. That is the guard against the two drifting apart later.

On the SPL sweep, I took your first option rather than approximate it: its smoothing
happens inside a Lanczos resample of linear amplitude, which the finished curve no longer
contains, so such a capture stays **unsmoothable** (calibration still switches). Doing it
properly needs the raw spectrum plus the SPL offset stored, as you suggested — worth its
own change rather than bolting onto this one. In practice the RTA is the case that
matters here; a relative sweep already keeps its raw form and is unaffected.

Also reverted the `MagnitudeAveraging` option I had added to `OverlayMath` in `b8c8abb`:
with smoothing delegated to the DSP layer it was dead, and the overlay display smoother is
untouched again.

App 684, DSP 880 green.

PR #66, opened 2026-08-02
Minimum-phase and excess group delay curves in the Group Delay mode

DIMOSUS:
## Summary

The Group Delay mode now renders the pair REW calls **minimum-phase group delay** and **excess group delay**, closing the curve half of the long-standing excess-delay follow-up.

- **Minimum Phase GD** (cyan, like the phase mode's Minimum Phase) is reconstructed from the gated magnitude alone — cepstral Bode relation via the existing `MinimumPhase.Reconstruct` — then pushed back through the **same** energy-weighted τ machinery (`τ = Re[T·conj(H)] / |H|²`), the same smoothing and the same validity gate as the measured curve. The reconstruction preserves |H| exactly, so both curves share one |H|² denominator and one gate: a bin is finite in every curve or NaN in every curve.
- **Excess GD** (green, like Excess Phase) is the pointwise measured − minimum. It stays **absolute** (referenced to the IR start), so it reads as the frequency-dependent arrival time of the all-pass part: a flat excess curve is a pure delay; deviations are crossovers/reflections that magnitude EQ cannot move.
- `GetGroupDelay` became a thin wrapper over the new `GetGroupDelayCurves`; the measured path is unchanged line-for-line and a pin test asserts the wrapper returns the set's Measured curve with no minimum-phase work paid for.
- Compare sources get all three curves (dashed/dimmed, same as the phase mode); the minimum/excess pair joins the Y-axis auto-fit for the main measurement only.
- Two new GDOpt checkboxes with explanatory tooltips; visibility persists through settings. A pre-feature settings file (fields absent) defaults the new curves **on**; a stored `false` survives the roundtrip.
- New `AnalysisCurveKind.MinimumPhaseGroupDelay` / `ExcessGroupDelay` are appended at the enum tail (the kind is persisted in overlay files); overlay capture/linking works via the stable keys `GroupDelay:MinimumPhaseGroupDelay:Main` / `GroupDelay:ExcessGroupDelay:Main`. GD mode is outside the overlays' amplitude space, so no psychoacoustic averaging applies to the new curves.

## Test plan

- `GroupDelayCurvesTests` (new, dsp): four synthetic discriminators + a wrapper pin —
  - pure delay δ(24 samples): minimum ≈ 0 (±1e-6 ms), excess = the delay;
  - one-pole 0.9ⁿ (minimum-phase by construction): excess ≈ 0 (±0.01 ms) while the system's own τ reaches ~0.19 ms;
  - second-order all-pass 1 kHz Q=2: the ~1.27 ms corner dispersion lands in excess, minimum stays flat (±0.05 ms);
  - differencer: NaN masks agree bin-exactly across all three curves;
  - `GetGroupDelay` ≡ `GetGroupDelayCurves(...).Measured`, with no minimum/excess computed by default.
- App tests: per-flag series presence/absence with independent toggles, Compare tagged min/excess curves, CurveTag labels, settings migration (default-on for legacy JSON, off-state roundtrip).
- Full suite green: 948 dsp + 772 app + 138 audio, 0 warnings (warnings-as-errors).

### DIFF
--- a/dsp/AnalysisModels.cs
+++ b/dsp/AnalysisModels.cs
@@ -25,7 +25,13 @@ public enum AnalysisCurveKind
     // no loopback reference behind it. Unlike a transfer function it carries an
     // absolute level, which is what makes a moving-microphone RTA average usable as
     // an equalization source.
-    InputSpectrum
+    InputSpectrum,
+    // The group-delay counterparts of MinimumPhase/ExcessPhase: the delay the
+    // gated magnitude alone implies (Bode relation), and the all-pass remainder
+    // measured − minimum that no minimum-phase equalizer can move. Appended at
+    // the end: the kind is persisted in overlay files.
+    MinimumPhaseGroupDelay,
+    ExcessGroupDelay
 }
 
 /// <summary>
@@ -60,3 +66,16 @@ public sealed record AnalysisCurve(
     string Name,
     IReadOnlyList<SignalPoint> Points,
     AnalysisCurveKind Kind = AnalysisCurveKind.Primary);
+
+/// <summary>
+/// The Group Delay mode's curve family, computed in one pass over one gate
+/// extraction: the measured group delay plus, when requested, the minimum-phase
+/// group delay implied by the gated magnitude alone and the excess
+/// (measured − minimum). The optional curves are null when the caller did not
+/// ask for the minimum-phase work. All present curves share one frequency grid
+/// and one validity gate: a bin is finite in every curve or NaN in every curve.
+/// </summary>
+public sealed record GroupDelayCurveSet(
+    AnalysisCurve Measured,
+    AnalysisCurve? Minimum,
+    AnalysisCurve? Excess);
--- a/dsp/DataHelper.Phase.cs
+++ b/dsp/DataHelper.Phase.cs
@@ -1057,7 +1057,37 @@ namespace Resonalyze.Dsp
             double plateauMs,
             double rightMs,
             double smoothingInverseOctaves,
-            double magnitudeGateDb = -30.0)
+            double magnitudeGateDb = -30.0) =>
+            GetGroupDelayCurves(
+                measurement,
+                gateOffsetMs,
+                leftMs,
+                plateauMs,
+                rightMs,
+                smoothingInverseOctaves,
+                magnitudeGateDb,
+                includeMinimumPhase: false).Measured;
+
+        /// <summary>
+        /// The Group Delay mode's curve family over ONE gate extraction: the
+        /// measured group delay and, when requested, the minimum-phase group
+        /// delay (from the gated magnitude via the Bode relation) plus the
+        /// excess (measured − minimum). All curves run through the same
+        /// energy-weighted τ evaluation and smoothing and share one validity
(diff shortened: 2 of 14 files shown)


DIMOSUS:
Review addressed in 281ad90 — all three points:

**Y auto-fit → measured only.** Agreed and reverted to the pre-feature semantics: `UpdateGroupDelayRange` now runs for the measured curve alone; minimum/excess render but clip off-scale like overlays. Your failure scenario is real — the local −40 dB validity gate tracks the rolloff's own envelope down to the −60 dB global backstop, so the band-edge slopes stay *valid* while the cepstral reconstruction reads them as ~1/f tens of ms. A new app test (`GroupDelay_AxisAutoFit_ReadsTheMeasuredCurveOnly`) pins the axis range as identical with the pair shown and hidden (it fails against the previous commit). One consequence to know: with the measured curve hidden and only min/excess shown, the axis falls back to the default −5…+5 window (pan/zoom within ±30 still works) — same as the old "nothing shown" behavior.

**Gate-geometry tests added.**
- `BulkDelayedMinimumPhaseSystem_SplitsDelayFromDispersion`: one-pole 0.9ⁿ behind a 5 ms bulk delay, gate offset on the arrival, 2 ms left Tukey shoulder (extractionStart = 144 ≠ 0). Excess reads flat 5 ms (±0.02) across 100 Hz–18 kHz; the one-pole dispersion (~0.19 ms toward DC) stays in the minimum curve.
- `WrapGate_LeftShoulderBeforeIrStart_KeepsTheSplit`: peak at sample 5 with a 48-sample left shoulder → negative extractionStart, cyclic-tail wrap. Measured and excess read the true absolute arrival (±1e-6 ms), minimum stays at zero — the time re-reference reaches exactly the curves it should.

**Stale Compare comment** rewritten: under Auto the gate *placement* is per-curve (resolved right above); only the length and smoothing are shared.

Suites: 950 dsp + 773 app + 138 audio, green, 0 warnings.

DIMOSUS:
Both cases fixed in 9342816, using exactly the proposed scheme:

- `UpdateGroupDelayRange(curves.Measured, …)` now runs when **measured OR excess** is shown — with measured hidden, the axis still follows the measured absolute level, which is where the in-band excess lives (their difference is the minimum curve, ≈ 0 away from the rolloffs), so an 8–10 ms excess-only layout stays on screen.
- Showing the minimum curve extends the fitted range with `min(minimum, 0) / max(maximum, 0)` (only when `hasValidData`), so at a 5 ms arrival the window becomes −2…7.2 ms and the ≈ 0 minimum curve is visible instead of clipped.
- Minimum alone keeps the −5…+5 default, which already contains it.
- The pair's own point values still never enter the fit — the band-edge cepstral swings remain excluded.

The axis test was the masked-by-geometry case you called out: it now uses a **240-sample (~5.44 ms) arrival** instead of the 1.45 ms delta and runs all four layouts with your suggested asserts (`Minimum < 0.2` / `Maximum > 5.0` for the all-three case, measured-level follow for excess-only, no zero-pin without the minimum curve, default window for minimum-only). Against the previous commit the first two layouts fail; with the fix all suites are green: 950 dsp + 773 app + 138 audio, 0 warnings.

PR #71, opened 2026-08-09
EQ Wizard: build the filter bank by hand, with shelves and undo

DIMOSUS:
The PEQ bank stops being a fixed rack of 32 mostly-greyed strips and becomes a
list the user builds: filters are added one at a time, reordered and thrown away
by dragging, undone with Ctrl+Z, kept between sessions, and can now be shelves as
well as bells.

## The bank

* Opens empty. Filters are added from a tile split into one zone per shape — PK,
  HS, LS — each with its own dashed frame, plus and label. The tile trails the
  last strip and disappears at 32.
* A strip is dragged by its **number** to reorder it (the bank re-orders live
  under the pointer) and dragged **out of the bank** to remove it, with the
  cursor turning into a bin outside and Escape putting the strip back.
* Right-clicking the number switches a filter between a bell and a shelf in
  place, keeping its frequency, Q and gain.
* A strip's index IS its filter number, its cell and the number an exported
  profile writes — so order is state, not decoration.

## Undo

Snapshots of the whole bank (filters, order, preamp), 100 deep. Field and fader
edits coalesce into one step once the bank goes quiet; an add, a removal, a
reorder, an import and a whole Auto Tune result are each a single step. The
source, the target and the Auto Tune settings are deliberately **not** part of
it, and the shortcut is bound at the panel — which costs a numeric field its own
in-place Ctrl+Z, a deliberate trade.

## Persistence

`EqWizardSettings` now carries the bank itself (frequency, Q, gain and shape per
filter, in slot order) plus the preamp, written once per undo step. Schema goes
9 → 10; a 7..9 file carries only the filter count and still opens, rebuilding the
ISO-centred spread those versions showed. An empty bank round-trips as empty —
only a missing list means "rebuild from the count".

## Shelving filters

`PeqBand` gains a `Type` as a defaulted fourth positional parameter, so no
existing call site changed and old files read back as bells. All realization goes
through one new dispatcher, `PeqBiquad`, which the response preview, the Virtual
DSP chain and the coefficient exports now share.

Shelves are parameterised by an **RBJ shelf Q** — the parameter the export
targets actually read:

| Format | Shelves |
| --- | --- |
| Equalizer APO, REW | `LSC` / `HSC` with a Q (verified against the APO configuration reference) |
| CamillaDSP | `Lowshelf` / `Highshelf`, same freq/gain/q triple |
| Generic CSV | new `Type` column; files without it still read |
| miniDSP, GraphicEQ | free — coefficients and a sampled curve carry any shape |
| EasyEffects | **cannot state ours** — declared unable, warned about, dropped |

EasyEffects' shelving bands are LSP filters whose steepness comes from a mode and
a slope multiplier rather than an RBJ Q, so writing our number into one would
produce a differently-shaped shelf on the target. The panel says exactly how many
filters would be left out and asks; the coordinator strips them either way, so a
file can never contain a shelf the reader would realize differently.

Import gains shelves as a side effect — a REW file's shelves used to be dropped
silently. The `LS 6dB` / `HS 12dB` and `LSC 10.8 dB` spellings state a corner
frequency or a dB-per-octave slope, so they are still skipped rather than read
into a shelf that would sit somewhere else.

Two consequences that would otherwise be silent bugs:

* **The Q conventions no longer restate a shelf.** They translate a bandwidth
  between half-gain points; a shelf has none (its Q is a knee), so scaling it by
  the gain would have printed a shelf that overshoots where the designed one does
  not.
* **The tuning sheets print shelves in their own table**, after the bells and
  keeping the bank numbering, because frequency, Q and direction do not mean what
  the three bell rows say.

Auto Tune is unchanged by design: bells only, and it still replaces the whole
bank, so a hand-dialled shelf does not survive a re-fit. That is now stated where
it is decided.

## Verification

1967 tests, zero warnings. New coverage: shelf math against the realized biquad
(full gain on its own side, unity on the other, exactly half at Fc, monotonic at
Q 0.707, overshoot at Q 3, analog prototype vs biquad), profile round-trips
keeping shape and order, the rejected parameterisations, export exclusion and
count, the sheet's table split with bank numbering, undo/redo semantics, the drag
hit test and settings round-trips.

Beyond the suites, the bank was driven end-to-end on a live panel in a scratchpad
harness (53 checks): grid cells and renumbering after every reorder, the add tile
appearing and disappearing at 32, the drag hit test against real strip geometry,
undo of a reorder/removal/type change, a curated bank surviving capture into a
fresh panel, and the actual curve — a +6 dB low shelf measuring **+5.94 dB at
30 Hz and 0.00 dB at 15 kHz**.

**Not machine-verified:** the mouse gestures themselves (the OLE drag loop, the
zone clicks, the bin cursor appearing) were exercised by hand, not by a test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,10 @@ the executable, nothing left on the machine — create an empty file named
 - EQ Wizard: equalize any measured frequency response — an impulse response from
   a file or History, a captured overlay slot, or a curve imported from text —
   toward its own target curve, designing an up-to-32-band parametric EQ (with
-  selectable microphone calibration), using a vertical fader bank, Auto Tune, a
+  selectable microphone calibration), using a vertical fader bank you add
+  filters — bells and low/high shelves — to one at a time and reorder or throw
+  away by dragging them, with undo/redo over the whole bank and the bank itself
+  kept between sessions, Auto Tune, a
   live results read-out, cross-tool PEQ import/export, and a printable
   tuning-sheet PDF. The car case it was built for is a **moving-microphone RTA in
   dB SPL**: no coherence, no impulse response, an absolute datum
@@ -340,9 +343,9 @@ the executable, nothing left on the machine — create an empty file named
       <h3>EQ Wizard</h3>
       <img src="assets/images/eq_wizard.png" alt="EQ Wizard mode">
       <p>Load an impulse response and design a parametric EQ toward its own target
-      curve with microphone calibration, Auto Tune, a 32-band fader bank, per-band
-      curves, a live results read-out, cross-tool import/export, and a
-      tuning-sheet PDF.</p>
+      curve with microphone calibration, Auto Tune, a drag-ordered fader bank of
+      up to 32 filters, per-band curves, a live results read-out, cross-tool
+      import/export, and a tuning-sheet PDF.</p>
     </td>
   </tr>
   <tr>
--- a/dsp/CamillaDspYamlFormat.cs
+++ b/dsp/CamillaDspYamlFormat.cs
@@ -4,10 +4,12 @@ using YamlDotNet.Serialization;
 namespace Resonalyze.Dsp;
 
 /// <summary>
-/// CamillaDSP config (YAML). Exports each PEQ band as a Biquad/Peaking filter plus a
-/// Gain filter for the preamp, wired into a two-channel pipeline. Imports Peaking
-/// biquads and the first Gain filter; other filter types are skipped. Peaking filters
-/// sum commutatively, so filter order does not affect the result.
+/// CamillaDSP config (YAML). Exports each PEQ band as a Biquad — Peaking, Lowshelf
+/// or Highshelf, all of which take the same freq/gain/q triple — plus a Gain filter
+/// for the preamp, wired into a two-channel pipeline. Imports those three biquad
+/// types and the first Gain filter; anything else is skipped. The filters are
+/// cascaded, and a cascade of biquads commutes, so filter order does not affect the
+/// result.
 /// </summary>
 public sealed class CamillaDspYamlFormat : IEqProfileFormat
 {
@@ -38,13 +40,15 @@ public sealed class CamillaDspYamlFormat : IEqProfileFormat
         for (int i = 0; i < curve.Bands.Count; i++)
         {
             PeqBand band = curve.Bands[i];
-            string key = $"peaking_{i:000}";
+            // The key names the slot, not the shape: a filter that changed type
+            // between two exports keeps its place in the pipeline list.
+            string key = $"band_{i:000}";
             filters[key] = new Dictionary<string, object?>
             {
                 ["type"] = "Biquad",
                 ["parameters"] = new Dictionary<string, object?>
                 {
-                    ["type"] = "Peaking",
+                    ["type"] = FilterTypeName(band.Type),
                     ["freq"] = EqTextNumbers.Format(band.FrequencyHz, "0.###"),
                     ["q"] = EqTextNumbers.Format(band.Q, "0.0"),
                     ["gain"] = EqTextNumbers.Format(band.GainDb, "0.0")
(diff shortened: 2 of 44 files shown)


DIMOSUS:
All four findings confirmed against the code and fixed in `3d62131`.

**1 (P1) — text tuning sheet printed shelves as `PK`.** Real, and the worst of the
four: the PDF took the shape from the band but `VirtualCrossoverSheet.FormatText`
spelled the keyword in place. The line now asks `PeqTextFile.TypeToken` for it, so
the sheet and the exported APO profile cannot drift apart again. Pinned by
`FormatText_NamesTheFilterShapeOfEveryBand` (PK/LSC/HSC) and
`FormatText_LeavesAShelfQAloneUnderAProportionalConvention`, which also checks the
bell beside the shelf is still restated.

**2 (P2) — last edit lost on exit.** Real. Fixed one level up rather than in the
panel: `FlushMeasurementSettings` now asks the panel to land anything still inside
the coalescing pause before the saver runs, which covers the shutdown fast path
and the docked settings hosts as well, since they all flush through it.

**3 (P2) — redo cleared by a narrowed gain range.** Real. `ApplyHistoryState` now
adopts `CaptureBankState()` — what the strips actually hold after clamping — as the
baseline, instead of the state that was requested.

**4 (P3) — unknown `PeqBandType` read three ways.** Real and worth the fix: the
realization and the header said bell, the shelf-dropping export and the sheet's
table split said shelf. There is now one predicate, `PeqBandTypes.IsShelving()`,
used by `PeqBiquad`, `PeqQConventions`, the export coordinator and `PdfSheet`, and
a band whose shape nothing recognises is normalised to a bell as it enters from the
settings file.

Verification: 1971 tests green, zero warnings. New pins for 1 and 4;
`AShapeNoMemberMatchesIsAcceptedByTheFileAndNormalisedOnLoad` covers the file half
of 4, and `AnUnknownShapeIsABellToEveryConsumerAlike` asserts the three consumers
agree.

2 and 3 are panel-level and outside the non-UI test convention, so they were
verified in the scratchpad harness that drives a live `EqWizardPanel` by
reflection — a fresh edit raises no settings change, `CommitPendingBankEdit()`
raises exactly one and the captured bank carries the new value; and after
narrowing Min Gain to −6 dB, undo clamps −12 → −6, redo is still offered and
returns to −2 dB. Both checks fail on the pre-fix code.

DIMOSUS:
The remaining P2 is real and is fixed in `0fa6357`.

The previous fix committed the bank but not the fields it is read from:
`DarkNumericUpDown` holds typed text until Leave or Enter, so
`CommitPendingBankEdit()` was still reading the pre-typing `Value`. The ordinary
close hid it — disabling the form takes focus out of the field, which commits the
text on the way — and the `WindowsShutDown` branch, which is the case the fix
existed for, does not. `CommitPendingBankEdit()` now lands the editors first (the
preamp and each strip's three fields via the existing public `CommitText()`, so
unparseable text still falls back to the committed value) and commits the bank
second.

`PeqSlotControlCommitTests` covers the case asked for — fields typed into and
never left — plus an untouched strip and text that does not parse. End-to-end
through the panel it is also in the scratchpad harness: typing alone leaves the
value at 777 Hz, `CommitPendingBankEdit()` lands 315 Hz and the −7.5 dB preamp,
and the captured settings carry them.

**On "zero warnings" — you are right and I was imprecise.** My figure came from
`dotnet build` of the solution locally, which reports 0. CI does emit two
`NETSDK1187`, and I should have said so rather than quoting a local number as if
it were the CI's. For the record on where they come from: they name
`NetSparkleUpdater.UI.WinForms` (a locale-cased resource in the package), this
branch changes no project, props or package file (`git diff main -- '*.csproj'
'Directory.*' 'global.json'` is empty), and the same warning appears five times in
the last main-branch run (31301805107). So they are pre-existing and third-party —
but they are not zero, and the accurate claim is "no warnings from this code, two
third-party SDK warnings on a clean CI restore".

1974 tests green.

Issue #86, opened 2026-08-18
Hello from autosound-tuning-skill: a Helix bank format (PR ready), hardware-verified Helix notes, and a file-level bridge with REW / TCC


ayukhno:
*A note on how this was written: this text, and the branch it points to, were produced by an AI coding agent working under my direction and review — the same way, judging by your `AGENTS.md`, much of Resonalyze is written. So it is written to be read comfortably by whichever agent you point at it as much as by you: file and member names are exact, every claim names its evidence, and every "we can" is something I have actually authorised.*

*Contact, in case a conversation is easier than issues: **ayukhno@gmail.com**, or name a messenger and how to reach you there.*

Hi — I run [autosound-tuning-skill](https://github.com/ayukhno/autosound-tuning-skill) (MIT code, CC BY-SA docs): a REW-driven car-audio tuning method written for AI coding agents, with a Python `rew_tool/` (joint-phase summation, an excess-phase EQ-boost gate, crossover realisation against acoustic targets, a versioned DSP-state ledger). Its desktop front-end, [Autosound TCC](https://github.com/ayukhno/autosound-tcc) (Apache-2.0, Qt, installs on macOS and Windows 11), is in development: it reads REW live over its API, shows the whole rig (crossovers, delays, gains, per-channel EQ) and puts the AI in a panel; it never writes to the processor. Same doctrine as yours — loopback-referenced sweeps on one time base for phase and timing, moving-mic RTA for magnitude — and the same target hardware (Helix DSP Ultra S is our reference processor).

I read `dsp/` end to end (and built it on macOS arm64 — the 1089 tests pass there, which is a nice property of the split you kept). The `AlignmentSelection` tie-breaks, the sum-loss objective with the dip penalty, and above all the register of *refuted* ideas in `TODO.md` are the most useful field notes on multi-way alignment I have seen in the open. Thank you for writing the "why" down.

A few things we can offer, roughly in order of how ready they are:

### 1. PR ready: the Audiotec-Fischer "Full EQ (30 bands)" bank as an `IEqProfileFormat`

[...]

PR #87, opened 2026-08-19
Add the Audiotec-Fischer 30-band bank as an EQ profile format

ayukhno:
Item 1 of #86, opened at your request there.

The HELIX / MATCH / BRAX DSP PC-Tool imports one channel's EQ from a tab-separated "Full EQ (30 bands)" bank — the same block REW exports with its Audiotec Fischer equaliser selected. Until now a wizard profile could only reach those processors through REW. This adds `AudiotecFischerFormat` to `EqProfileFormats.All`, both directions:

- **Export** writes PK bells and the LS_Q / HS_Q shelves (a centre frequency and a Q — the shelf this library realizes, so they travel unchanged), always exactly 30 slots with unused ones as `None`, and REW's row shapes byte for byte: a bell row ends in the empty TargetT60 cell, a shelf row at its Q, an unused row at its `None` — because that is the layout the PC-Tool is known to accept. REW's Bandwidth column is written as Fc / Q (the RBJ relation, consistent with the `PeqQConvention` crib).
- **Import** reads the same; Q wins over the bandwidth column unless the Q cell is blank (then Q = Fc / BW); a shelf without a Q comes in at `PeqTextFile.DefaultShelfQ`. AP1 / AP2 rows move phase only, which no `PeqBand` can state, so they are skipped and never written; `Enabled False` rows are skipped like OFF filters; a BOM, CR/LF, ragged rows, trailing tabs and spaces in place of tabs are tolerated; an all-`None` bank is recognised as a valid empty profile.
- **What the layout cannot carry, said rather than hidden.** The bank has no place for the preamp — the channel gain is a separate PC-Tool control — so this adds `IEqProfileFormat.CarriesPreamp` (default `true`, the same shape as `SupportsShelvingFilters`): the format declares `false`, writes no preamp, reads it back as 0, and the shared round-trip test asserts that by declaration. More than 30 bands is refused on export with an `ArgumentException` (which `EqWizardImportExportCoordinator` already surfaces) rather than truncated; `AudiotecFischerFormat.SlotCount` is public for a later device profile.

Tests: `AudiotecFischerFormatTests` pins import against a real REW export of a 20-band channel (18 bells + both shelves + ten empty slots), the row shapes on export, the preamp/slot-budget/all-pass rules, tolerance, recognition vs. foreign text, and round trips; `EqProfileShelfFormatsTests.RoundTrip_KeepsEachBandsShapeAndOrder` now runs over this format too.

README's import/export paragraph lists the format. `TODO.md`'s "device export needs a target-device profile" item keeps the residual: the EQ Wizard does not yet warn about a non-zero preamp left behind the way it warns about a dropped shelf, and the miniDSP / generic paths still have no budget or gain profile. I can follow up on the warning; it is app-side, and I cannot run `Resonalyze.App.Tests` here.

Verified on macOS arm64: `Resonalyze.Dsp.Tests` 1104 / 1104 (was 1089), and `source/Resonalyze.sln` compiles with `-p:EnableWindowsTargeting=true` — the App and Audio suites need your Windows CI. No warnings, no suppressions.

Background: the fixture and the format knowledge come from [autosound-tuning-skill](https://github.com/ayukhno/autosound-tuning-skill), where the same bank is parsed and generated in Python (`rew_tool/atf_eq.py`) for Helix tunes driven from REW; the C# here is written fresh against this repo's conventions and is offered under the repository's MIT licence.

### DIFF
--- a/dsp/AudiotecFischerFormat.cs
+++ b/dsp/AudiotecFischerFormat.cs
@@ -71,6 +71,10 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
     private const string LowShelfType = "LS_Q";
     private const string HighShelfType = "HS_Q";
     private const string EmptyType = "None";
+    // The slots that legitimately hold no magnitude band: the two all-pass types move
+    // phase only, which no PeqBand can state. Anything else claiming to be an active
+    // filter must be readable, or the bank is not this file.
+    private static readonly string[] PhaseOnlyTypes = ["AP1", "AP2"];
 
     public string Name => "Audiotec Fischer 30-band bank";
     public string Extension => "txt";
@@ -151,7 +155,8 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
                 continue;
             }
 
-            if (!TryParseSlot(line, out int slotNumber, out PeqBand? band))
+            SlotKind kind = ReadSlot(line, out int slotNumber, out PeqBand band);
+            if (kind == SlotKind.NotASlot)
             {
                 // Not a slot row at all: this is not the bank's table.
                 tableIntact = false;
@@ -159,13 +164,16 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
             }
 
             slotsSeen++;
+            // A slot that claims to hold a filter but cannot be read is a band the
+            // import would silently drop from a fixed table — refuse the file instead.
+            tableIntact &= kind != SlotKind.Unreadable;
             // The bank is a fixed table: slot n is the nth row. A file that skips,
             // repeats or renumbers rows is not the channel's slot table, whatever
             // its header says.
             tableIntact &= slotNumber == slotsSeen && slotsSeen <= SlotCount;
-            if (band.HasValue)
+            if (kind == SlotKind.Band)
             {
-                bands.Add(band.Value);
+                bands.Add(band);
             }
         }
 
@@ -194,28 +202,50 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
             fields[3].Equals("Type", StringComparison.OrdinalIgnoreCase);
     }
 
-    // Reads one slot row: whether the line IS a slot of this bank (its number), and
-    // the band it holds, if any. Empty (None), all-pass, disabled, unknown-type and
-    // malformed rows are slots holding no band — a bank is a fixed table, so most
-    // rows of a real file are expected to be empty. Telling "no band here" from
-    // "not this bank's table" is what lets an incomplete file be refused.
-    private static bool TryParseSlot(string line, out int slotNumber, out PeqBand? band)
+    /// <summary>What one line of a bank turns out to be.</summary>
+    private enum SlotKind
     {
-        band = null;
-        slotNumber = 0;
+        /// <summary>Not a row of this table at all.</summary>
+        NotASlot,
+
+        /// <summary>A slot that legitimately holds no band: None, all-pass, or OFF.</summary>
+        Empty,
+
+        /// <summary>A slot holding a band this profile can state.</summary>
+        Band,
+
+        /// <summary>A slot claiming a filter whose type or numbers cannot be read.</summary>
+        Unreadable
+    }
+
+    // Reads one slot row: whether the line IS a slot of this bank (its number), what
+    // kind of slot, and the band it holds. Telling "no band here" from "not this
+    // bank's table" is what lets an incomplete file be refused; telling both from
+    // "a filter I cannot read" is what stops a band going missing unnoticed.
+    private static SlotKind ReadSlot(string line, out int slotNumber, out PeqBand band)
+    {
+        band = default;
 
         string[] fields = SplitRow(line);
         if (fields.Length < 4 ||
             !int.TryParse(fields[0], NumberStyles.Integer, CultureInfo.InvariantCulture, out slotNumber) ||
             fields[3].Length == 0)
         {
-            return false;
+            slotNumber = 0;
+            return SlotKind.NotASlot;
         }
 
-        if (fields.Length < 6 ||
-            fields[1].Equals("False", StringComparison.OrdinalIgnoreCase))
+        // An OFF filter occupies its slot and contributes nothing, whatever it says
+        // afterwards — as elsewhere in these formats.
+        if (fields[1].Equals("False", StringComparison.OrdinalIgnoreCase))
         {
-            return true;
+            return SlotKind.Empty;
+        }
+
+        if (fields[3].Equals(EmptyType, StringComparison.OrdinalIgnoreCase) ||
+            PhaseOnlyTypes.Contains(fields[3], StringComparer.OrdinalIgnoreCase))
+        {
+            return SlotKind.Empty;
         }
 
         PeqBandType type;
@@ -234,14 +264,17 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
         }
         else
         {
-            // None, AP1, AP2 or anything else this layout may grow: a slot, no band.
-            return true;
+            // An enabled slot of a type this reader does not know. It may well be a
+            // filter that shapes magnitude, so passing it off as an empty slot would
+            // quietly change the tune.
+            return SlotKind.Unreadable;
         }
 
-        if (!EqTextNumbers.TryParse(fields[4], out double frequencyHz) ||
+        if (fields.Length < 6 ||
+            !EqTextNumbers.TryParse(fields[4], out double frequencyHz) ||
             !EqTextNumbers.TryParse(fields[5], out double gainDb))
         {
-            return true;
+            return SlotKind.Unreadable;
         }
 
         // The Q cell may be blank on a hand-edited row: a bell can still be read
@@ -253,7 +286,7 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
                 if (!EqTextNumbers.TryParse(FieldAt(fields, 7), out double bandwidthHz) ||
                     bandwidthHz <= 0)
                 {
-                    return true;
+                    return SlotKind.Unreadable;
                 }
 
                 q = frequencyHz / bandwidthHz;
@@ -268,11 +301,11 @@ public sealed class AudiotecFischerFormat : IEqProfileFormat
             !double.IsFinite(q) || q <= 0 ||
             !double.IsFinite(gainDb))
         {
-            return true;
+            return SlotKind.Unreadable;
         }
 
         band = new PeqBand(frequencyHz, q, gainDb, type);
-        return true;
+        return SlotKind.Band;
     }
 
     // Tab-separated as exported; a copy pasted through something that turned the
--- a/dsp/AudiotecFischerFormat.cs
+++ b/dsp/AudiotecFischerFormat.cs
@@ -1,0 +1,253 @@
+using System.Globalization;
+using System.Text;
+
+namespace Resonalyze.Dsp;
+
+/// <summary>
+/// The Audiotec-Fischer "Full EQ (30 bands)" bank: the tab-separated block REW
+/// exports with its Audiotec Fischer equaliser selected, and the file the HELIX /
+/// MATCH / BRAX DSP PC-Tool imports into one channel's EQ:
+/// <code>
+/// Audiotec_Fischer_Full_EQ_(30_bands)
+/// Number	Enabled	Control	Type	Frequency(Hz)	Gain(dB)	Q	Bandwidth(Hz)	TargetT60(ms)
+/// 1	True	Auto	PK	80.0	-2.0	4.00	20
+/// 2	True	Auto	LS_Q	50.0	-1.0	0.70
+/// 3	True	Auto	None
+/// </code>
+/// The bank is the channel's slot table, so it always holds exactly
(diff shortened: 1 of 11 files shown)


dsp/IEqProfileFormat.cs
```diff
@@ -30,6 +30,19 @@ public interface IEqProfileFormat
...
+    /// Whether the layout has a place for the whole-profile gain
+    /// (<see cref="EqualizationCurve.PreampDb"/>). False means the target keeps that
+    /// gain in a control the file never reaches — a car DSP's channel gain — so an
+    /// export leaves the preamp out and an import reads it as 0; the caller is
+    /// expected to say so, as it does for a dropped shelf.
+    /// </summary>
+    /// <remarks>
+    /// Defaults to true: the formats that carry a preamp line, an output-gain field
+    /// or a gain biquad realize it by construction.
+    /// </remarks>
```
DIMOSUS, line 43: Could we wire `CarriesPreamp` into the EQ Wizard export path in this PR rather than leave it only as format metadata? Right now Audiotec Fischer correctly returns `false`, but `EqWizardImportExportCoordinator` only handles `SupportsShelvingFilters`. A curve with e.g. `PreampDb = -6` therefore exports successfully with the -6 dB silently omitted, so the tune imported into PC-Tool no longer matches what Resonalyze showed. Please add a warning/confirmation analogous to the dropped-shelf warning, ideally telling the user which channel gain must be entered manually.

dsp/AudiotecFischerFormat.cs
```diff
@@ -0,0 +1,253 @@
...
+        string[] fields = SplitRow(line);
+        if (fields.Length < 6 ||
+            !int.TryParse(fields[0], NumberStyles.Integer, CultureInfo.InvariantCulture, out _) ||
+            fields[1].Equals("False", StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        PeqBandType type;
+        if (fields[3].Equals(BellType, StringComparison.OrdinalIgnoreCase))
```
DIMOSUS: REW's Audiotec Fischer equalizer can also contain `Modal` filters. Semantically those are still peaking EQ filters with Fc/Gain/Q; the modal-specific T60 field is optimizer metadata and does not prevent representing the resulting filter as `PeqBandType.Peaking`. With the current type switch they are silently dropped. Could you support the exact Modal token REW writes as a peaking band and add an import test using a representative REW row?

dsp/AudiotecFischerFormat.cs
```diff
@@ -0,0 +1,253 @@
...
+    public bool TryImport(string text, out EqualizationCurve curve)
+    {
+        ArgumentNullException.ThrowIfNull(text);
+
+        bool recognized = false;
+        var bands = new List<PeqBand>();
+
+        foreach (string rawLine in text.TrimStart('\uFEFF').Split('\n'))
+        {
+            if (bands.Count >= EqualizationCurve.MaxBandCount)
```
DIMOSUS: This should be tied to the format's fixed `SlotCount`, not `EqualizationCurve.MaxBandCount` (32). More importantly, `recognized` currently becomes true from the bank/column header alone, so a truncated file can import successfully as an empty curve and overwrite the user's current EQ. Conversely a malformed bank with 31–32 active rows can import, but the same format then refuses to export it. Could you validate the bank structure against the 30-slot layout (while still allowing `None`, AP and disabled rows to count as slots) and reject incomplete/out-of-range banks rather than treating them as a successful empty/oversized import?

DIMOSUS (changes requested):
Thanks — the implementation is clean and the test coverage is substantial. I found three cases I'd like to see addressed before merging because they can silently change or erase a user's EQ: preamp loss on export, REW Modal filters being dropped on import, and overly permissive recognition of incomplete/non-30-slot banks.

ayukhno:
Thanks — all three are real ways to lose a user's tune, and all three are addressed in cd4f220.

**1. `CarriesPreamp` wired into the wizard.** `EqWizardExportTarget` passes it through (the tuning sheet carries the preamp, so it reports true), `EqWizardImportExportCoordinator.PreampDroppedBy(target, curve)` reports the gain an export would leave behind — zero when the format carries it or the curve has none — and `EqWizardPanel` warns before writing, next to the shelf warning: it names the dB and says to enter it in the channel's own gain control on the device. `EqWizardPreampExportTests` covers the coordinator surface (including the tuning sheet warning for neither, and the bank exporting the bands but not the −6.5 dB); it needs your Windows CI, as before.

**2. Modal rows.** Imported as `PeqBandType.Peaking` and written back as `PK` — the processor has no modal slot and the T60 is what the optimizer aimed at, not part of the filter it runs. One correction to the premise, from measuring it here: REW's modal filter has **no Q of its own** — the API refuses `q` for it (`"Modal filters do not have a q"`) and reports `frequency` / `gaindB` / `t60Target`. Its realized width is not any textbook T60 identity either: a modal filter set to 120 Hz / −6 dB / T60 300 ms in REW 5.40 realizes **Q 11.6** (fitted from REW's own EQ response, 0.014 dB rms; the `π·f0·T60/ln(1000)` identity would say 16.4). So the Q column is the only thing worth reading, which is what the import does, and the test row is that filter.

If you would rather the test pin a byte-exact REW export of a bank containing a modal row, I can add one — I have REW here and can produce it.

**3. Recognition and the slot budget.** `TryImport` now validates the bank as the table it is: exactly `SlotCount` rows, numbered in order, with `None` / `AP1` / `AP2` / disabled / malformed rows counting as slots but yielding no band. A truncated, renumbered or over-long file is no longer recognised, so it can no longer arrive as an "empty bank" and replace the EQ on screen — and the budget is the format's own `SlotCount`, not `EqualizationCurve.MaxBandCount`, so what imports is what this format can export. Parsing stays lenient about everything that does not change the table's shape (BOM, CR/LF, ragged rows, spaces for tabs, a blank Q). New tests: the truncated / header-only / 31-row / renumbered cases, and a full 30-band bank that round-trips.

`Resonalyze.Dsp.Tests` 1107/1107 on macOS arm64; the solution compiles with `-p:EnableWindowsTargeting=true`. README's import/export paragraph and the `TODO.md` residual are updated (the residual is now the miniDSP / generic device profile, not this bank).

ayukhno:
Follow-up on the modal row (1149069): the test now imports a byte-exact bank instead of my reconstruction. REW 5.40 Beta 132, this equaliser selected, two modal filters and a bell:

```
Audiotec_Fischer_Full_EQ_(30_bands)
Number	Enabled	Control	Type	Frequency(Hz)	Gain(dB)	Q	Bandwidth(Hz)	TargetT60(ms)	
1	True	Auto	Modal	120.0	-6.0	11.59	10.35	300	
2	True	Auto	Modal	45.0	-4.5	9.48	4.75	600	
3	True	Auto	PK	200.0	-2.0	4.00	50.00	
```

So: the token is `Modal`, the Q column carries the bell's own Q, the bandwidth column stays `Fc / Q`, and the T60 cell is an integer (the row is otherwise a bell row). You were right that the row is representable — the only correction to the premise is that the modal filter has no Q of its own in REW's model (the API refuses `q` for it and reports `frequency` / `gaindB` / `t60Target`), and the Q it writes here is derived: for these two rows it lands on `π·f0·T60/ln(1000) · 10^(gain/40)` to the printed precision (16.37·0.708 = 11.59; 12.28·0.773 = 9.48 — the second factor being the same RBJ↔Symmetric scale as in `PeqQConventions`). Measuring REW's own EQ response for the 120 Hz row gives Q 11.59 as well, so what the column states is what the filter realizes.

Nothing to derive from T60 then: the import reads the Q column like any bell row, which is what the code does.

source/Tools/Eq/EqWizardPanel.cs
```diff
@@ -851,6 +852,34 @@ private bool ConfirmShelvingBandsDropped(
...
+        if (preampDb == 0)
+        {
+            return true;
+        }
+
+        string gain = FormattableString.Invariant($"{preampDb:+0.0;-0.0} dB");
+        return MessageBox.Show(
+            FindForm(),
+            $"{target.Name} has no place for the preamp, so the {gain} would be left " +
+            "out and the exported bands alone are that much louder than the tune on " +
```
DIMOSUS: Small correctness issue in the warning text: `louder` is only true for a negative preamp. The UI allows positive preamp values too; with `+6 dB`, dropping it makes the exported result 6 dB quieter, not louder. Could you either make this wording sign-aware (`louder` / `quieter`) or use neutral wording such as `would differ from the tune on screen by that gain`? The `Enter {gain}` instruction itself is correct.

dsp/AudiotecFischerFormat.cs
```diff
@@ -0,0 +1,287 @@
...
+        else
+        {
+            // None, AP1, AP2 or anything else this layout may grow: a slot, no band.
+            return true;
+        }
+
+        if (!EqTextNumbers.TryParse(fields[4], out double frequencyHz) ||
+            !EqTextNumbers.TryParse(fields[5], out double gainDb))
+        {
+            return true;
```
DIMOSUS: Could we distinguish a valid slot that intentionally carries no `PeqBand` from a malformed *active* EQ slot? Right now a recognized PK/Modal/LS_Q/HS_Q row with an invalid Fc/Gain (and similarly an unusable Q/BW below) returns `true` with `band = null`, so a complete 30-row bank still imports successfully but silently loses that filter. For this fixed hardware-bank format I'd prefer malformed active EQ rows to invalidate the bank, while `None`, AP1/AP2 and `Enabled=False` remain valid slots with no band. I would also avoid treating arbitrary unknown active types as silently empty unless they are explicitly known non-magnitude slot types.

DIMOSUS (commented):
Two follow-up findings from checking the fixes. The original three requested changes look addressed; these are new edge cases introduced/exposed by the updated paths.

DIMOSUS:
Sorry for such a nitpicky review — I don't cut my own PRs any slack either and run them through several review agents as well.

ayukhno:
Both addressed in af74323 — and no need to apologise for the review: the second pass found the better rule.

**The warning's wording.** You are right, the sign was baked in. It now reads "louder" for a dropped cut and "quieter" for a dropped boost; the `Enter {gain}` instruction is unchanged.

**Malformed active slots.** Agreed, and this is the rule the format should have had: in a fixed table there is nowhere for a band the reader failed to state, so dropping one leaves a changed tune with nothing to show for it. `ReadSlot` now answers with a kind — `NotASlot` / `Empty` / `Band` / `Unreadable` — and the bank is refused when any enabled slot is `Unreadable`, i.e.

- a band type (`PK`, `Modal`, `LS_Q`, `HS_Q`) whose centre, gain or width cannot be read — including a bell with neither a Q nor a bandwidth, and a zero width;
- an enabled type this reader does not model at all (`Notch`, say). Only `None` and the all-pass slots are known-empty now, from an explicit list rather than "anything unrecognised".

`Enabled False` still makes a slot empty whatever it carries, as OFF filters do elsewhere in these formats, and the leniency that does not change the table's shape (BOM, CR/LF, ragged rows, spaces for tabs, a blank Q with a usable bandwidth, a shelf without a Q) is untouched.

Tests are split accordingly: `Import_SkipsTheSlotsThatHoldNoBand` for the three empty kinds, `Import_RefusesASlotThatClaimsAFilterItCannotRead` for the five refusals plus the same unmodelled row turned OFF. `Resonalyze.Dsp.Tests` 1108/1108 on macOS arm64; the solution compiles with `-p:EnableWindowsTargeting=true`; README updated with the exception to "malformed entries are skipped".

DIMOSUS (commented):
One last robustness issue in the format recognition path.

dsp/AudiotecFischerFormat.cs
```diff
@@ -0,0 +1,320 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(text);
+
+        bool headerSeen = false;
+        int slotsSeen = 0;
+        bool tableIntact = true;
+        var bands = new List<PeqBand>();
+
+        foreach (string rawLine in text.TrimStart('\uFEFF').Split('\n'))
+        {
```
DIMOSUS, line 144: `headerSeen` is still set by either the actual bank header or the generic column header. That means `Number / Enabled / Control / Type ...` plus 30 well-numbered rows is accepted even if `Audiotec_Fischer_Full_EQ_(30_bands)` is missing entirely. Since this importer now intentionally uses strict fixed-bank recognition, could we require the real `BankHeader` explicitly and treat the column header as optional/auxiliary only? A regression test for `column header + 30 None rows, but no BankHeader => false` would pin it.

ayukhno:
Fixed in 7526354.

You are right that the recognition was still loose in the one place it matters: REW writes that same `Number / Enabled / Control / Type ...` line above every equaliser's table, so it says nothing about whose thirty rows follow — a bank from a different processor with the same column shape would have imported here and replaced the EQ on screen. Only `Audiotec_Fischer_Full_EQ_(30_bands)` grants recognition now; the column line is skipped when present and never required, so a copy that arrives without it still reads.

Tests: the case you asked for (column header + 30 `None` rows, no bank header → `false`), plus the two directions that must keep working — the bank header alone importing its table, and the byte-exact REW export, which carries both.

`Resonalyze.Dsp.Tests` 1109/1109 on macOS arm64.

DIMOSUS (approved):

Issue #90, opened 2026-08-19
Spec: importing a REW loopback-referenced impulse response (item 5 of #86)


ayukhno:
Item 5 of #86, written down first as you asked: what a REW measurement carries, how much of it survives each export route, and what it would map to in `ImpulseResponseFile`. Everything below was measured against a live REW 5.40 Beta 132 (API 0.9.6) on 2026-08-19, not read off a forum; the files behind every claim are published (CC BY 4.0) in [autosound-measurements](https://github.com/ayukhno/autosound-measurements) — [`rew-export-samples/`](https://github.com/ayukhno/autosound-measurements/tree/main/cars/vw-passat-b8-sedan-lhd/2026-06-15_front-set-01/rew-export-samples) has the same channel in all three forms with a README, and the eight-channel set next to it is already in v7.

## 1. What REW has

A sweep measured with **`timingReference: Loopback`** and **`timingOffset: 0`** is on the same absolute base yours is: t = 0 is the loopback arrival, and the IR's own `delay` is the tract's delay. (The other references — `Acoustic`, `None`, or any non-zero timing offset — are not that base; see §5.)

Per measurement REW exposes: `startTime` (the time of sample 0, seconds), `sampleRate`, `timingReference`, `timingOffset`, `delay` (its own interpolated arrival), `startFreq` / `endFreq`, the sweep length, `splOffsetdB`, `signalToNoisedB`, notes, uuid, date. There is **no coherence** for a sweep, no level-meter snapshot, no audio-session record — the optional parts of v7 simply stay absent.

## 2. Two traps worth putting in the spec before anything else

**The IR endpoint peak-normalises by default.** `GET /measurements/{id}/impulse-response` returns every IR with its peak at exactly ±1.0 (`unit: percent` — of the peak, not of full scale), so two channels carry **no level relation at all**. `?normalised=false` gives the raw IR in percent of full scale (÷100 → fraction of FS). Our own cross-check harness summed peak-normalised IRs for a day before this surfaced; anything doing sum-loss or gain balance on REW data must ask for the raw form.

[...]

PR #93, opened 2026-08-20
Import a REW loopback-referenced impulse response (#90)

ayukhno:
Implements #90 — the REW → Resonalyze import path, in the shape you asked for: the text export first, a live HTTP client left out.

## What it does

A `.txt` alongside the `.json` and `.wav` the Load dialog already takes. The reader lives in `Resonalyze.Dsp` so the part worth testing is testable on any platform; the app side is thin wiring.

**Why the text export and not the WAV.** Of REW's three routes out, it is the only one that states the time of sample 0. A WAV carries the samples and no timing unless t = 0 is pinned to a whole sample on the way out — which cannot express a fractional zero, and which rewrites the measurement's own start time to make the cut (measured: one of our eight channels moved by 0.033 sample when exported that way).

**The fractional zero.** REW anchors its buffer on the microphone peak, an integer sample, and lets the loopback arrival fall where it falls: for the eight channels of the set behind this work, `-startTime · fs` came out as 95072.000, 95541.000, 95522.000, 95720.967, 95604.082, 95713.793, 95599.417, 95629.073. Rounding that away costs up to half a sample — 5 µs at 96 kHz — which is the resolution your arrival estimator works at. So `FractionalSampleShift.AdvanceCircular` moves the whole part as a rotation and the fractional part as a linear phase ramp, which for a band-limited signal is not an approximation but the exact interpolation. It is circular on purpose: the pre-roll of a deconvolved sweep holds the harmonic images that belong at negative time, and they wrap to the tail rather than being cut.

**What is refused, and why it has to be.** Two header lines decide whether a file is usable at all and neither shows in the samples: an export made with *normalise* on has had its peak scaled to one and holds no level relation to any other channel, and an export with the IR window applied is a view of the response rather than the response. The minimum-phase version is refused too — its excess phase is the one thing this import exists to keep. A file written without headers is refused because the headers are where the time base is.

Whether the sweep was loopback-referenced is **read but not enforced in the DSP layer** — that is a routing decision, so the reader exposes `IsLoopbackReferenced` and the app refuses on it, quoting REW's own excitation line back to the user.

**What is assumed, and said out loud rather than assumed silently.** The import posts a notice naming: the bit depth (a text file of fractions of full scale keeps no trace of it), the playback channel (REW does not say which output it played through, so `Mono` rather than a side it never carried), and — only when the header lacks them — the swept band and the sweep's length. The sweep length is read from the excitation line's `512k` where present, because REW keeps the impulse response shorter than the sweep that made it and the harmonic geometry follows the sweep. The notice also states what the format cannot carry at all: no coherence, no level meters, no SPL anchor, and no microphone calibration — REW applies that to its own curves, not to the impulse response, so an imported IR is uncalibrated whatever REW showed.

The result lands as a **measurement**, like a recorded sweep and unlike a loaded file: nothing on disk holds it in this program's terms, so it must not become a save target pointing back at the user's REW export.

## Checking

Against a real 262144-sample export of a car measurement, read here and compared with the same measurement converted from REW's API by a separate implementation in another language: **max |difference| 1.9e-10** on both the raw buffer and the re-referenced one — float32 text rounding, 5.6e-8 of peak — and the same transfer peak index.

`Resonalyze.Dsp.Tests` 1117/1117 on macOS arm64; the solution builds clean with `-p:EnableWindowsTargeting=true`.

The fixtures behind every timing claim are published (CC BY 4.0) next to the dataset, if you want to check the reader against REW's own bytes rather than against my description:
https://github.com/ayukhno/autosound-measurements/tree/main/cars/vw-passat-b8-sedan-lhd/2026-06-15_front-set-01/rew-export-samples

## What I have not done

**The UI path has not been run yet** — the logic it leans on is covered in the Dsp suite, but no one has actually clicked Load on a `.txt` in a running Resonalyze. I have a Windows VM now and will exercise it before this is asked to be merged; say if you would rather I do that first and you review afterwards.

Two decisions I would rather you make than guess:

1. **Non-loopback exports are refused outright.** #90 raised importing them as `RecordedSweep` instead — shape real, position its own. I left that out because it adds a second meaning to one menu item, but it is a small change if you want it.
2. **Protective high-passes.** A file-based import has no capture moment at which to remove them, so nothing is de-embedded here. Asking on import (a per-channel `ProtectiveHighPassConfiguration`) or accepting a declaration alongside the file are both possible; our own published sets carry the declaration per channel.

Reverse bridge (Resonalyze → REW) next, investigated before it is written, as agreed in #90.

### DIFF
--- a/dsp/RewImpulseResponseTextFile.cs
+++ b/dsp/RewImpulseResponseTextFile.cs
@@ -63,6 +63,15 @@ public sealed class RewImpulseResponseTextFile
     public int PeakIndex { get; private init; }
 
     /// <summary>
+    /// The arrival this header implies, in samples: the buffer's peak minus the
+    /// reference. For a loopback-referenced sweep it is the tract's delay, and it is
+    /// positive by physics — the microphone cannot hear the sweep before the reference
+    /// does. A file where it is not is on some other time base; see the refusal in
+    /// <see cref="TryParse"/>.
+    /// </summary>
+    public double ImpliedArrivalSamples => PeakIndex - TimeZeroIndex;
+
+    /// <summary>
     /// REW's peak before normalisation — <em>interpolated</em>, so it sits a little above
     /// the largest sample in the data (0.002 dB on the file this reader was written
     /// against). It is the sub-sample peak, not one of the numbers below it.
@@ -277,10 +286,29 @@ public sealed class RewImpulseResponseTextFile
         }
 
         double timeZero = -startTime * sampleRate;
+        int peakIndexValue = values.TryGetValue("Peak index", out double peakIndex)
+            ? (int)Math.Round(peakIndex)
+            : 0;
         if (!(timeZero >= 0) || timeZero >= samples.Count)
         {
             problem = $"t = 0 falls at sample {timeZero:0.###} of {samples.Count}, outside the " +
                 "buffer: these samples do not contain the reference arrival";
+            return false;
+        }
+
+        // REW's text export does not record a timing offset — the header of a
+        // measurement taken with one is word for word the header of one taken without,
+        // and the offset is simply baked into the start time. What it cannot hide is
+        // the consequence: a large enough offset pushes t = 0 past the buffer's peak,
+        // and an arrival before the reference is not something a loopback measurement
+        // can produce. Measured: a 4 ms offset on a 2.7 ms arrival puts t = 0 127
+        // samples after the peak, which REW's own API reports as a delay of -1.32 ms.
+        if (values.ContainsKey("Peak index") && peakIndexValue <= timeZero)
+        {
+            problem = $"the header puts the reference at sample {timeZero:0.###} and the " +
+                $"buffer's peak at {peakIndexValue}, so the arrival would precede the " +
+                "reference — which a loopback-referenced sweep cannot do. The usual cause " +
+                "is a timing offset applied in REW, which this export does not record";
             return false;
         }
 
@@ -291,9 +319,7 @@ public sealed class RewImpulseResponseTextFile
             SampleRate = sampleRate,
             SampleIntervalSeconds = interval,
             StartTimeSeconds = startTime,
-            PeakIndex = values.TryGetValue("Peak index", out double peakIndex)
-                ? (int)Math.Round(peakIndex)
-                : 0,
+            PeakIndex = peakIndexValue,
             PeakValueBeforeNormalisation =
                 values.TryGetValue("Peak value before normalisation", out double peak) ? peak : 0.0,
             DataOffsetDb = values.TryGetValue("Data offset (dB)", out double offset) ? offset : 0.0,
--- a/dsp/RewImpulseResponseTextFile.cs
+++ b/dsp/RewImpulseResponseTextFile.cs
@@ -316,7 +316,8 @@ public sealed class RewImpulseResponseTextFile
     {
         if (!TryParse(text, out RewImpulseResponseTextFile? file, out string? problem) || file == null)
         {
-            throw new FormatException($"This REW impulse-response export cannot be imported: {problem}.");
+            throw new FormatException(
+                $"This REW impulse-response export cannot be imported — {problem}.");
         }
 
         return file;
--- a/dsp/RewImpulseResponseTextFile.cs
+++ b/dsp/RewImpulseResponseTextFile.cs
@@ -358,19 +358,63 @@ public sealed class RewImpulseResponseTextFile
         }
     }
 
-    // The band line is written with the exporting machine's thousands separator, so
-    // "19,999.9" and "19.999,9" are the same frequency on two different machines.
-    private static bool TryReadGrouped(string token, out double value) =>
-        double.TryParse(
-            token.Trim(),
-            NumberStyles.Float | NumberStyles.AllowThousands,
+    /// <summary>
+    /// Reads a number written with either grouping convention — "19,999.9" and
+    /// "19.999,9" are the same frequency on two different machines.
+    /// </summary>
+    /// <remarks>
+    /// The file was written somewhere else, so neither the invariant culture nor the
+    /// one this program is running under is the right authority: trying them in turn
+    /// gets "20,1" wrong twice over, once as 201 under a culture that groups by comma
+    /// and once by refusing it. The token itself says which character is the decimal
+    /// point — the rightmost of the two when both appear, and otherwise the only one,
+    /// unless it stands three digits from the end and alone, which is how a thousands
+    /// group looks in both conventions.
+    /// </remarks>
+    private static bool TryReadGrouped(string token, out double value)
+    {
+        value = 0;
+        string trimmed = token.Trim();
+        if (trimmed.Length == 0)
+        {
+            return false;
+        }
+
+        int lastDot = trimmed.LastIndexOf('.');
+        int lastComma = trimmed.LastIndexOf(',');
+        char? decimalSeparator;
+        if (lastDot >= 0 && lastComma >= 0)
+        {
+            decimalSeparator = lastDot > lastComma ? '.' : ',';
+        }
+        else if (lastDot >= 0 || lastComma >= 0)
+        {
+            char only = lastDot >= 0 ? '.' : ',';
+            int at = lastDot >= 0 ? lastDot : lastComma;
+            int occurrences = trimmed.Count(character => character == only);
+            bool looksGrouped = occurrences > 1 ||
+                (occurrences == 1 && trimmed.Length - at - 1 == 3);
+            decimalSeparator = looksGrouped ? null : only;
+        }
+        else
+        {
+            decimalSeparator = null;
+        }
+
+        string normalized = decimalSeparator is char separator
+            ? trimmed
+                .Replace(separator == '.' ? "," : ".", string.Empty, StringComparison.Ordinal)
+                .Replace(separator, '.')
+            : trimmed
+                .Replace(".", string.Empty, StringComparison.Ordinal)
+                .Replace(",", string.Empty, StringComparison.Ordinal);
+
+        return double.TryParse(
+            normalized,
+            NumberStyles.Float,
             CultureInfo.InvariantCulture,
-            out value) ||
-        double.TryParse(
-            token.Trim(),
-            NumberStyles.Float | NumberStyles.AllowThousands,
-            CultureInfo.CurrentCulture,
             out value);
+    }
 
     // "512k Log Swept Sine, 1 sweep at -10.0 dBFS using a loopback as a timing reference"
     private static (int? Length, int? Count) ReadExcitation(string? excitation)
--- a/dsp/RewImpulseResponseTextFile.cs
+++ b/dsp/RewImpulseResponseTextFile.cs
@@ -1,0 +1,412 @@
+using System.Globalization;
+
+namespace Resonalyze.Dsp;
+
+/// <summary>
+/// REW's <c>File → Export → Impulse response as text</c>, read back.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Of the ways REW can hand an impulse response to another program, this is the only one
+/// that carries the absolute time base: the header states the time of sample 0, so a
+/// measurement taken against a loopback reference can be put back on the base it was
+/// measured on. A WAV export carries the samples and no timing at all unless t = 0 is
+/// pinned to a whole sample on the way out — which cannot state a fractional zero, and
+/// which rewrites the measurement's own start time to make the cut.
+/// </para>
+/// <para>
+/// Two things in the header decide whether the file is usable, and both have a safe
+/// default that is not the one wanted here. An export made with <em>normalise</em> on has
+/// its peak scaled to one, so it holds no level relation to any other channel; an export
+/// with the IR window applied is not the impulse response but a view of it. Both are
+/// refused rather than read, because neither announces itself in the samples.
+/// </para>
+/// <para>
+/// The samples are fractions of full scale — the same numbers REW's API serves for
+/// <c>?normalised=false</c>, to float precision.
+/// </para>
+/// </remarks>
+public sealed class RewImpulseResponseTextFile
+{
+    private const string FileMarker = "Impulse Response data saved by REW";
+    private const string DataStartMarker = "* Data start";
+
+    private const string NoHeaders =
+        "not a REW impulse-response text export, or one written without its headers " +
+        "(the headers carry the time base, so a file without them cannot be placed in time)";
+
+    private RewImpulseResponseTextFile(double[] samples)
+    {
+        Samples = samples;
+    }
+
+    /// <summary>The impulse response as REW served it, as a fraction of full scale.</summary>
+    public double[] Samples { get; }
+
+    public int SampleRate { get; private init; }
+
+    public double SampleIntervalSeconds { get; private init; }
+
+    /// <summary>
+    /// The time of sample 0, negative for every loopback measurement: REW anchors the
+    /// buffer on the microphone peak and lets the reference arrival fall where it falls.
+    /// </summary>
+    public double StartTimeSeconds { get; private init; }
+
+    /// <summary>
+    /// Where t = 0 — the loopback arrival — sits in this buffer, in samples. Fractional
+    /// in general: REW pins the peak to a whole sample, not the reference.
+    /// </summary>
+    public double TimeZeroIndex => -StartTimeSeconds * SampleRate;
+
+    /// <summary>The largest sample's index, as REW states it.</summary>
+    public int PeakIndex { get; private init; }
+
+    /// <summary>
+    /// REW's peak before normalisation — <em>interpolated</em>, so it sits a little above
+    /// the largest sample in the data (0.002 dB on the file this reader was written
+    /// against). It is the sub-sample peak, not one of the numbers below it.
+    /// </summary>
+    public double PeakValueBeforeNormalisation { get; private init; }
+
+    /// <summary>The SPL offset REW would add to turn these samples into dB SPL.</summary>
+    public double DataOffsetDb { get; private init; }
+
+    /// <summary>The measurement's name in REW, if the header carried one.</summary>
+    public string? MeasurementName { get; private init; }
+
+    /// <summary>The capture device line, if the header carried one.</summary>
+    public string? Source { get; private init; }
+
+    /// <summary>The excitation line verbatim — sweep length, count, level and reference.</summary>
+    public string? Excitation { get; private init; }
+
+    /// <summary>
+    /// The swept band REW reports. Best effort: the header writes it in the exporting
+    /// machine's number format, so a file that cannot be read here still parses and
+    /// leaves this null rather than failing over metadata.
+    /// </summary>
+    public double? LowFrequencyHz { get; private init; }
+
+    public double? HighFrequencyHz { get; private init; }
+
+    /// <summary>
+    /// The sweep's length in samples, from the excitation line's <c>512k</c> / <c>1M</c>
+    /// label. The impulse response is often shorter than the sweep that produced it, so
+    /// this is the honest source for a sweep duration; null when the label is unfamiliar.
+    /// </summary>
+    public int? SweepLengthSamples { get; private init; }
+
+    /// <summary>How many sweeps REW averaged, from the excitation line.</summary>
+    public int? SweepCount { get; private init; }
+
+    /// <summary>
+    /// The impulse response re-referenced so that <b>sample 0 is the loopback arrival</b>,
+    /// which is the convention a transfer impulse response is stated in here. The whole
+    /// part of the offset is a rotation, the fractional part an exact shift; REW's
+    /// pre-roll wraps to the tail, where the harmonic images of a swept measurement
+    /// belong.
+    /// </summary>
+    public double[] ToLoopbackReferencedImpulseResponse() =>
+        FractionalSampleShift.AdvanceCircular(Samples, TimeZeroIndex);
+
+    /// <summary>
+    /// Reads an export, or explains in <paramref name="problem"/> why this one cannot be
+    /// trusted on the time base it claims.
+    /// </summary>
+    public static bool TryParse(string text, out RewImpulseResponseTextFile? file, out string? problem)
+    {
+        ArgumentNullException.ThrowIfNull(text);
+        file = null;
+        problem = null;
+
+        bool markerSeen = false;
+        bool dataStarted = false;
+        string? measurement = null;
+        string? source = null;
+        string? excitation = null;
+        double? low = null;
+        double? high = null;
+        var values = new Dictionary<string, double>(StringComparer.OrdinalIgnoreCase);
+        var samples = new List<double>();
+
+        foreach (string rawLine in text.TrimStart('\uFEFF').Split('\n'))
+        {
+            string line = rawLine.Trim();
+            if (line.Length == 0)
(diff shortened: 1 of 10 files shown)


dsp/RewImpulseResponseTextFile.cs
```diff
@@ -0,0 +1,412 @@
...
+        return true;
+    }
+
+    /// <summary>
+    /// Reads an export, throwing with the reason when it cannot be trusted.
+    /// </summary>
+    public static RewImpulseResponseTextFile Parse(string text)
+    {
+        if (!TryParse(text, out RewImpulseResponseTextFile? file, out string? problem) || file == null)
+        {
```
DIMOSUS, line 361: **P1:** This only proves that REW used a loopback timing reference; it does not prove that the timing offset was zero. #90 explicitly calls out `timingOffset != 0` as a different time base that must not be marked `SynchronizedLoopback`, and the implementation plan repeated that requirement. A non-zero offset would therefore be silently baked into `StartTimeSeconds` and imported as a real path delay. The text export does not appear to expose `timingOffset` as a separate header field, so this may need an explicit import/design decision rather than another string check.

source/Shell/Form1.FileOperations.cs
```diff
@@ -187,6 +197,148 @@ private async Task LoadImpulseResponseFileAsync(string path)
...
+                (string.IsNullOrWhiteSpace(file.Excitation)
+                    ? string.Empty
+                    : $" (REW says: \u201c{file.Excitation}\u201d)") +
+                ". Its shape is real, but nothing ties its zero to anything outside its " +
+                "own measurement, so it cannot be placed on this session's time base. " +
+                "Measure it in REW with a loopback as the timing reference to import it.");
+        }
+
+        double[] samples = file.Samples;
+        double[] referenced = await Task.Run(file.ToLoopbackReferencedImpulseResponse);
```
DIMOSUS: **P2:** The agreed import behavior said to refuse a REW file whose sample rate does not match the current measurement session. This path currently passes `file.SampleRate` straight into `RestoreImpulseResponse` without comparing it to the configured rate. Afterward `ApplyLoadedImpulseResponseState()` reconfigures controllers from the current Measurement Settings, so a 96 kHz import into a 48 kHz session can leave the measurement and controller/session configuration describing different rates. Please reject the mismatch explicitly with a useful message, as `ImportRecordedSweep` already does.

DIMOSUS (commented):
I found four issues that I think are worth fixing before merge.

source/Shell/Form1.FileOperations.cs
```diff
@@ -187,6 +197,148 @@ private async Task LoadImpulseResponseFileAsync(string path)
             panel => panel.RefreshSplAvailability());
     }
 
+    // A measurement made in REW, brought over on the time base it was measured on.
+    //
+    // REW's text export is the only one of its routes that states the time of sample 0,
+    // so a sweep it measured against a loopback can be placed on the same absolute base
+    // a measurement taken here sits on — which is the whole point: two programs' results
```
DIMOSUS, line 312: **P2:** This import does not claim `expSweepMeasurement` while the file is being read/parsed and the fractional FFT is running. During those awaits the Record button can start a sweep; then this import reaches `RestoreImpulseResponse`, sees `InProgress`, and fails. `ImportRecordedSweepAsync` already uses `expSweepMeasurement.Claim()` specifically to close this race across file decoding/analysis. The REW path should have equivalent lifetime protection (with the restore path able to publish while holding its own claim).

dsp/RewImpulseResponseTextFile.cs
```diff
@@ -0,0 +1,412 @@
...
+    private static bool Says(string note, string head, string tail) =>
+        note.StartsWith(head, StringComparison.OrdinalIgnoreCase) &&
+        note.Contains(tail, StringComparison.OrdinalIgnoreCase);
+
+    private static void ReadBand(string text, out double? low, out double? high)
+    {
+        low = null;
+        high = null;
+        string trimmed = text.Replace("Hz", string.Empty, StringComparison.OrdinalIgnoreCase).Trim();
+        int to = trimmed.IndexOf(" to ", StringComparison.OrdinalIgnoreCase);
```
DIMOSUS, line 389: **P2:** This is not actually portable between the locale that exported the REW file and the locale running Resonalyze. `InvariantCulture` followed by `CurrentCulture` only covers those two cultures, and can also misread an ambiguous token such as `20,1` under invariant parsing (comma treated as a thousands separator). Please parse REW's grouped-number syntax explicitly (`1,234.5` vs `1.234,5`) and add tests where the file's number format differs from `CurrentCulture`.

DIMOSUS (changes requested):

ayukhno:
The UI path has now been run, on Windows 11 arm64, against a real REW export — so the caveat in the description is retired. Six checks, all as intended, and two things worth fixing that only showed up by looking at the dialogs.

**Import.** `w_L.ir-text.txt` (262144 samples, 96 kHz, `Start time` −0.9952187499999998 s) imports, and the notice reads back the right numbers: the loopback reference at sample 95541 of REW's buffer, now sample 0 of the transfer response. The impulse plot puts the arrival at sample ~460 against the 459 that the same measurement carries in the published v7 file — 4.79 ms, matching REW's own 4.7788 ms delay.

**Refusals**, each from the same real export with exactly one header line altered, so nothing differs but the reason to refuse: normalised → refused; IR window applied → refused; acoustic timing reference → refused, quoting REW's excitation line back (`512k Log Swept Sine, 1 sweep at -10.0 dBFS using the acoustic timing reference`). No exceptions, no half-imported state.

**Round trip.** The imported measurement saves as v7 and loads back without complaint.

## Two fixes from seeing it (21be9c7)

1. **The notice named what the format cannot carry but not what the import assumed in its place** — bit depth and playback channel. Those are precisely the quiet decisions the notice exists to prevent, so they are stated now, with the reason they are harmless: neither touches the samples, both describe the sweep the result is filed under.
2. The refusals read `cannot be imported: the export is normalised: its peak has been scaled to one` — two colons in one sentence, because the wrapper and the reason each wanted one. The wrapper takes a dash instead.

`Resonalyze.Dsp.Tests` 1117/1117 on macOS arm64; the solution builds clean with `-p:EnableWindowsTargeting=true`.

The two decisions from the description are still yours: whether a non-loopback export should import as `RecordedSweep` rather than be refused, and whether protective high-passes should be asked for on import.

source/Measurements/ExpSweepMeasurement.cs
```diff
@@ -463,11 +463,29 @@ public void RestoreImpulseResponse(
...
-                    "Cannot load an impulse response while a measurement is running.");
+                if (inProgress && !claimed)
+                {
+                    throw new InvalidOperationException(
+                        "Cannot load an impulse response while a measurement is running.");
+                }
+
+                if (!inProgress)
+                {
+                    inProgress = true;
```
DIMOSUS, line 483: **P1:** This makes every `RestoreImpulseResponse` call fail before it can restore anything. We set `inProgress = true` here (or enter with an external `Claim()` already holding it), then later call the public `Init(...)`; `Init` immediately checks `InProgress` and throws `Cannot reinitialize an active measurement.` The current CI confirms this directly: App Tests fail broadly through `RestoreImpulseResponse -> Init` (63 failures). This path needs to initialize through the claim-aware/core path (e.g. `InitCore`) rather than the guarded public `Init`, with coverage for both a direct restore and a restore while an external claim is held.

DIMOSUS (commented):
One new blocker in the race fix; the other three addressed review items look good.

ayukhno:
All four addressed — f342ca4 and 031b7b0. The timing-offset one turned into an experiment, so I will start there.

## P1 — the timing offset

You were right that the text export does not expose it, and right that it needed a decision rather than another string check. I measured what it actually looks like: same channel, same session, exported once normally and once with **`Timing offset = 4 ms`** set in the measurement dialog.

The header is **word for word identical**. Same excitation line — `512k Log Swept Sine, 1 sweep at -9.0 dBFS using a loopback as a timing reference` — same fields, same everything. The offset is simply baked into the start time. There is nothing in this format to test for.

What it cannot hide is the consequence:

```
96000                    // Peak index
-1.0013229166666666      // Start time (seconds)
```

That start time puts t = 0 at sample **96127** — a hundred and twenty-seven samples *after* the buffer's peak. REW's own API reports that measurement's `delay` as **−1.32 ms**, which is the same 127 samples. Sound does not reach the microphone before it reaches the loopback.

So the import now **refuses a file whose peak does not follow its reference**, naming a timing offset as the usual cause. And the honest limit is stated rather than left to be discovered: an offset *smaller* than the arrival keeps the peak after t = 0 and is indistinguishable from a shorter path. The notice therefore reports the arrival the header implies, so a user who knows what REW showed for that measurement can catch what the format cannot.

One consequence worth noting for later: the **API route does expose `timingOffset` directly**. That is now a concrete argument for the live-import path being worth having, beyond convenience — it can check something the file cannot.

## P2 — the claim

Claimed before the read and held until the result is published. That needed `RestoreImpulseResponse` to tell a claim from a run, which is the shape `ImportRecordedSweep` already uses: a run still blocks, a claim does not, and with nothing held it takes its own for the publish. The v7 load path gets the same protection as a side effect, since it went through the same guard.

## P2 — the sample rate

Refused now, with the same shape of message the recorded-sweep path uses, compared against `measurementSettings.Measurement.BuildConfiguration().Signal.SampleRate` before anything is restored.

## P2 — the numbers

Agreed, and worse than you put it: invariant-then-current does not merely fail to cover the case, it reads `20,1` as **201** on the first attempt and never reaches the second. Parsing now decides from the token itself — the rightmost of the two characters when both appear, otherwise the only one, unless it stands three digits from the end and alone, which is what a thousands group looks like in both conventions. Tests cover both conventions under two different `CurrentCulture` values, the grouped case, and a band line that cannot be read at all, which leaves the band absent rather than failing the file.

---

`Resonalyze.Dsp.Tests` **1119/1119** on macOS arm64; the solution builds clean with `-p:EnableWindowsTargeting=true`. The UI path was run again after the earlier round and is unchanged in behaviour apart from the new refusals.

The synthetic fixtures needed fixing along the way, and the way they broke is worth a line: they placed the pulse *on* the reference, so the new check refused them — an arrival of exactly zero is not physical either. They now sit 34 samples in against a reference at 20.25, which also makes the fractional-shift assertion read a real 13.75-sample arrival rather than a rotation onto sample zero.

dsp/RewImpulseResponseTextFile.cs
DIMOSUS: This is useful as an additional sanity check, but it still does not close the original requirement. The new code explicitly accepts a non-zero timing offset whenever it is smaller than the physical arrival, because that case remains indistinguishable from a shorter path in this text format. We would still stamp such a result as `SynchronizedLoopback`, even though #90 says any `timingOffset != 0` is a different time base and must not be treated as synchronized. Since the format cannot prove zero offset, I think this still needs an explicit trust/design boundary (for example, user confirmation/assertion that timing offset was zero, or a route that carries the metadata) rather than treating the negative-arrival heuristic as sufficient validation.

DIMOSUS:
Everything else from my review is resolved now. The only remaining open point is the **P1 around `timingOffset`**: REW's text export still cannot prove that the measurement was made with `timingOffset == 0`, while any non-zero offset puts it on a different time base and should not be marked `SynchronizedLoopback`. Once we have an explicit trust/confirmation boundary for that case, I think my review is clear.

ayukhno:
CI is green again (70c0d60). The regression was mine: `RestoreImpulseResponse` set `inProgress` and then
called the guarded public `Init`, so every restore threw — the plain v7 load included, since it takes its own
claim. It goes through `InitCore` now, with tests for a direct restore, a restore under an externally held
claim, and a refused restore giving back the claim it took.

Reading every caller of `Claim()` afterwards turned up a second thing the same commit got wrong: the REW
import scoped its claim `using var`, holding it to the end of the method, so `ApplyLoadedImpulseResponseState`
redrew with `InProgress` still true — and `RefreshCurrentModePlot` gates `includeCurves` on `!InProgress`, so a
successful import drew the plot *without* the curve it had just imported. The modal notice then held the claim
for as long as it was open. `ImportRecordedSweepAsync` had already met this and says so in a comment; this path
copied the claim and not the scope. Fixed by closing the block right after the publish.

## The timing offset

You are right, and I will stop defending the check as validation. It catches an offset large enough to pull
t = 0 past the peak and nothing else; a smaller one is still stamped `SynchronizedLoopback`, which is what #90
forbids.

Rather than argue the boundary, I measured what the metadata route gives, using the REW API client we already
had for our own dataset work.

**The same measurement, both ways.** This is the 4 ms export from the last round — REW measurement #68,
`m-L (sw) timeOffset=4ms`, the very file whose header I said was word for word identical to an offset-free one.

| | text export | `GET /measurements/68/impulse-response` |
|---|---|---|
| `Start time` | `-1.0013229166666666` | `startTime` `-1.0013229166666666` |
| arrival | `-1.3229 ms`, from `Peak index` 96000 against t = 0 at 96127 | `delay` `-0.0013240` s |
| **the offset** | **nothing** | `timingOffset` **`0.004`**, `timingRefTime` `-0.004` |

Same time base, to the last digit. The one field that decides whether the result may be called synchronized
exists on one route and not the other.

**And across a session.** Ten measurements from 2026-08-20 — eight front, two rear taken on a different base:

| | `timingReference` | `timingOffset` | `timingRefTime` | `delay` |
|---|---|---|---|---|
| sw, w_L, w_R, m_L, m_R, tw_L, tw_R, c | `Loopback` | `0.0` | `±1.1e-16` | +2.91 … +9.67 ms |
| r_L / r_R | `Loopback` | `0.0077003` | `-0.0077003` | **−4.60 / −2.05 ms** |

Three things follow.

**The offset is exposed as a number, and redundantly.** `timingRefTime` is exactly `-timingOffset` on all
three distinct values seen so far — 0, 4 ms, 7.7003 ms — so the API route can read the offset and check it
against a second field.

**`timingReference` is not evidence of a zero offset.** It reads `Loopback` in every row above, including #68
at 4 ms and the rears at 7.7 ms. That was your point in round one; here are twelve rows of it.

**Our heuristic is the shadow of the offset, not a test for it.** #68 and the rears all have an offset larger
than their arrival, so `delay` goes negative and the negative-arrival check refuses them. That is why it works
when it works, and exactly why it is blind to an offset smaller than the arrival. It stays as a sanity check;
it is not the boundary.

**The text export still cannot prove zero.** `Start time` already has the offset folded in, and with
`Peak index` it yields the arrival and nothing else. (REW does spell the offset out in the measurement's
`notes` — "with 4.0000 ms … timing offset" — but notes are free text the user can edit, and they are not in
the export anyway.)

So the boundary I would build, split by route:

- **Text export (this PR).** The file cannot prove a zero offset, so it does not get to claim one. The import
  asks once, before publishing: was REW's Timing Offset zero for this measurement? **Yes** → stamped
  `SynchronizedLoopback`, with the notice recording that this was *asserted by the user*, not proven by the
  file. **No or unsure** → imported as `RecordedSweep`.
- **API route (separate, as agreed).** Reads `timingOffset` and stamps `SynchronizedLoopback` only when it is
  exactly zero, with `timingRefTime` as the cross-check. No question asked, because nothing is being assumed.
  This is no longer a hypothetical argument for that path — the field is there and populated. That route can
  also show REW's own wording back to the user in the import notice, since `/measurements/{id}` carries the
  `notes` REW writes ("with 4.0000 ms … timing offset") — useful as a human cross-check next to the number,
  never as the thing we gate on: notes are free text and a `PUT` replaces them.

**Why `RecordedSweep` is the honest fallback rather than a new state.** A timing offset shifts the whole axis
by a constant: it leaves everything *inside* the measurement intact and invalidates only comparison *between*
measurements — word for word what `RecordedSweep` already means in `TimingReference.cs`. The shape is real, a
reflection 8 ms after the direct sound is still 8 ms, and it cannot join Time Alignment.

Two things I would rather hear from you before writing it:

1. **A question, or a default?** The above adds one modal question to every text import. The alternative is no
   question — always `RecordedSweep` from a text file, and let the API route be the only way to earn
   `SynchronizedLoopback`. That is stricter, needs no new UI, and is arguably the more honest reading of #90.
   I lean towards it now that the metadata route is demonstrably real; the cost is that a text import can
   never join Time Alignment, however certain the user is.
2. **Is `RecordedSweep` acceptable as that fallback**, or would you rather refuse the import outright when a
   zero offset cannot be established?

Happy to share the raw pull for those ten measurements if the numbers are useful to you beyond this thread.

DIMOSUS:
This is a very useful result, and I think it gives us a better option than either treating every text import as `RecordedSweep` or merely asking the user to assert that the offset was zero.

If `timingOffset` is a constant shift of REW's time axis, then we should be able to **undo it**, rather than require it to be zero.

From your numbers, the relationship appears to be:

`physical arrival = (PeakTime + StartTime) + timingOffset`

or equivalently:

`correctedStartTime = StartTime + timingOffset`

So for the API route, a non-zero timing offset should not necessarily disqualify the measurement from `SynchronizedLoopback`: if the API tells us its value, we can remove it and restore the original loopback time base. `timingRefTime ≈ -timingOffset` is then a useful independent consistency check.

Before implementing that, could you verify it experimentally with the same measurement exported/measured at 0 ms and +4 ms? After removing the 4 ms offset, I would expect both to produce the same referenced IR and the same arrival, apart from numerical precision. That would give us a very strong regression fixture.

For the text route, I think the same model gives us a better fallback UX. Instead of asking only “was Timing Offset zero?”, ask for the **Timing Offset value used in REW**. If the user supplies it (including 0 ms), compensate it and import as `SynchronizedLoopback`. If they do not know it, import as `RecordedSweep`: the internal timing of the measurement is still valid, but its position relative to other measurements is not established.

I agree that the negative-arrival test should remain only as a sanity check, not as proof of the time base.

And yes, please share the raw API pulls for those measurements. The zero / 4 ms / 7.7003 ms cases would make excellent fixtures for both the current importer and the future API path.

ayukhno:
Both done — the experiment and the raw API pulls: https://github.com/ayukhno/autosound-measurements/releases/tag/rew-timing-offset-2026-08-23 (CC BY 4.0, full method and
numbers in the bundle's README).

Your model is confirmed **exactly** — but on the IR start, not the peak.

Six captures of one midrange, mic untouched: offsets 0 / 4 / 7.7003 ms, then a second
block 0 / 4 / 0. The extra zero-offset captures are the control; without them rig drift
and offset handling cannot be separated.

At 96 kHz a 4 ms offset is 384.0 samples — an integer. Every capture at an
integer-sample offset returns `timeOfIRStartSeconds + timingOffset` to the last digit:

```
#77 (0 ms)  #78 (4 ms)  #80 (0 ms)  #81 (4 ms)  ->  254.0000000000 samples, all four
```

Two deviations, both accounted for: **#79** at 7.7003 ms is 739.2288 samples, not an
integer, and the residual is exactly that fractional part — REW keeps the IR start on
the sample grid. **#82** moved 254 -> 253 at zero offset: one sample of rig drift.

## The peak will not do as an anchor — an argument for #107 from the measurement side

I ran this on the peak first and got a 3.3-6.7 us residual that looked systematic. It
was not the model. `timeOfIRPeakSeconds` wandered ~2.6 us per capture, and moved 3.6 ns
**between two reads of the same stored measurement** with no re-measurement, while
`timeOfIRStartSeconds` stayed bit-identical. Three zero-offset captures of one speaker:

```
IR start   254.0000   254.0000   253.0000
peak       260.1532   259.5590   258.9090
```

So a regression fixture should assert on the IR start, where the tolerance is zero. On
the peak it needs +/- 1 sample — loose enough to pass a broken implementation.

Related, for anything that measures drivers sequentially: the IR start drifted 1 sample
over 6 captures / 18 minutes, and the drift is per **capture**, not per clock — 17 idle
minutes moved it 0.5 us, consecutive captures ~2.6 us each.

## Two things I did not expect

**The text export rounds `Start time` to the sample grid** — -96124.000000 samples
against the API's -96124.166686 for the same measurement, 0.167 sample. Every other
field matches the API exactly, so it is rounding, not a different definition. #90 asks
for ~0.1 sample, so the export exceeds that budget before any offset arithmetic. Not an
argument against the text route — an argument that the two routes need different stated
error budgets.

**The offset is stated three times on the API and all three agree**: `timingOffset`,
`timingRefTime` (its exact negative, 2.2e-16 across all six), and in prose in `notes`,
at full precision. You already proposed the second as a check; the third is free.

I tested the obvious objection by hand-editing a note to claim 5.0000 ms while
`timingOffset` still read 0.004 — it is in the bundle as `meas-78-notes-hand-edited.json`.
A naive `with ([0-9.]+) ms` read is fooled; the structure check and the numeric
cross-check each reject it independently. The limit, plainly: editing **only** the number
keeps the structure intact, so the numeric field must stay authoritative and the notes
are a third vote, never a source.

## The text route

The offset reaches the export by no path at all. Notes never appear — three exports
taken while the notes were being edited were byte-identical. What looks like an offset
in my headers is only the measurement **title**, which I typed; renaming it changed that
one line and nothing else. Relying on the title would be the heuristic you rejected in
round one, since a rename would silently change the import.

So your plan for the text route stands unchanged: ask for the value, compensate if
given, `RecordedSweep` if not.

## One suggestion about order

#86 had the text route first and the API client separately, later. I think that is worth
reversing. The API is not a more convenient way to get the same thing — it is the only
route where #90's stated precision is reachable. The text route should exist, but it is
structurally second-class for timing. Your call; I mention it because the measurement
changed my own view of which is the main path.

## The branch now does what you proposed

Pushed as `6d15da2`, rebased onto current main — the #106 collision was two methods
landing in the same place in `Form1.FileOperations.cs` and nothing more. Your CI is green
on the merge commit: App 1301/1302 (1 skip), Audio 142/142, DSP 1209/1209, publish and
installer.

The question the import asks changed from "was it zero?" to "what was it?", and the
answer decides the reference:

- **a value, zero included** — a user assertion. It is taken back out and the result is
  stamped `SynchronizedLoopback` on that word.
- **"I don't know"** — not a failure but a different measurement. The shape is real, the
  position is not, which is exactly what `RecordedSweep` already means here.
- **Cancel** — the import is abandoned and nothing is said.

`RewImportTiming` decides it, beside the `TimingReference` enum and without a window, for
the same reason `SampleRateOptions.Resolve` lives where it does — it can then be tested
without one. Undoing the offset is a move of t = 0 and nothing else, so
`ToLoopbackReferencedImpulseResponse` takes it as an argument and the samples are
untouched. The sign is the measurement's, not an argument's: subtracting the offset from
the reference index is what reproduces the un-offset arrival to the last digit.

**The parse-time refusal of an arrival that precedes the reference is gone**, and that is
deliberate. It refused the one file the question exists to rescue — the 4 ms export whose
t = 0 sits 127 samples after its peak. The reader reports the implied arrival instead,
and a stated offset the file contradicts is refused with the offset that would work
named in the message.

The notice now says which of the two the import became, and on whose word: the arrival is
true on the user's, never on the file's.

## Still open

Only whether you want this split. I did the offset revision and the rebase in this PR
rather than wait, since you were mid-sprint and a conflicting branch is a poor thing to
hand a reviewer — say the word and I will lift the revision into its own PR.

DIMOSUS (approved):

DIMOSUS:
The API route looks like a great direction now that we know it carries the timing metadata the text export cannot preserve reliably. I’d keep the text importer as the offline/fallback path, and implement the REW API integration as a separate PR so it can be reviewed and validated on its own. In particular, the API path can use `timingOffset`/`timingRefTime` directly and avoid asking the user to reconstruct the time base manually.

PR #95, opened 2026-08-20
Keep the configured sample rate when the ASIO probe says nothing (#92)

### LINKED ISSUE #92: ASIO sample rate falls back to 44100 when the driver refuses a quick second probe, and the status label can disagree with the selection
Setting the ASIO sample rate to 96 kHz does not survive closing and reopening the settings window: it comes back as 44100, silently, while the driver's own control panel still shows 96 kHz. Pressing Apply from that state would persist 44100 — a measurement session at half the intended rate, with nothing on screen having said so.

Two separate things, one of which is visible in the screenshot below.

## 1. The rate list is rebuilt from a probe that is allowed to fail, and the selection falls back without saying so

`MeasurementOptions.cs`:

```csharp
int selectedSampleRate = availableRates.Contains(preferredSampleRate)
    ? preferredSampleRate
    : availableRates[0];
```

For ASIO, `availableRates` comes from `GetSupportedSampleRates()` → `asioDriverInfo.SupportedSampleRates`, and the comment right there already names the failure mode:

> From the last driver probe, not a fresh open: RefreshAsioDriverInfo always runs first, and a second open of the same driver moments later is what some drivers refuse (leaving an empty rate list).

Closing the settings window and reopening it immediately *is* a second open moments later. When that probe comes back empty or reduced, the saved 96000 is not in `availableRates`, so the selection silently becomes `availableRates[0]` — 44100 here — and the user's configuration is gone with no message. The failure is worst exactly when the driver is momentarily busy, which is also when the user is least likely to suspect the app.

Suggestion, in the order I would try them: when the probe returns nothing usable, keep the stored rate and leave the selection alone rather than rewriting it from an empty list; and when a fallback genuinely has to happen, show it (the same salmon-coloured status the error path uses) instead of swapping the number quietly. It might also be worth not persisting a rate that came from a fallback rather than from the user.

## 2. The status label can state a rate the combo does not show

With the settings window open, the sample-rate combo reads **96000** while the line under it reads **"44100 Hz supported"**, in green. `UpdateAsioStatusLabels` composes that text from two sources that can drift apart:

```csharp
int sampleRate = GetSelectedSampleRate();
labelAsioSampleRateStatus.Text = asioDriverInfo.SupportsSampleRate
    ? $"{sampleRate} Hz supported"
    : $"{sampleRate} Hz not supported";
```

— the number from the live selection, the verdict from whatever the last probe cached. Whichever of the two is stale, the sentence is confidently wrong, and it is the one line a user checks to confirm the rate took effect. (In my case it was the line that told me something was off, so it did its job by accident.)

## Environment

Resonalyze on Windows 11 arm64 in a Parallels VM on Apple silicon, Focusrite Scarlett 2i2 4th gen, Focusrite USB ASIO driver, 96 kHz set in Focusrite's own control panel. The virtual machine is presumably what makes the driver slow enough to refuse the quick second open — but the fallback and the label are the same on any machine whose driver ever fails a probe, and a busy ASIO device is not exotic.

## Repro

1. ASIO backend, Focusrite USB ASIO, sample rate 44100.
2. Set the sample rate to 96000 and Apply. The driver accepts it — Focusrite's tray panel confirms 96 kHz.
3. Close the settings window and reopen it immediately.
4. The rate reads 44100 again, while the driver is still at 96 kHz.

Happy to send a PR for both if you would like it in that shape — I have the branch open for #90 anyway. Say the word and I will keep it separate from that one.

ayukhno:
Fixes #92. Verified on the setup that produced it — Windows 11 arm64 in Parallels, Focusrite USB ASIO — and the repro no longer reproduces.

## What was wrong

Two faults that happened to sit on top of each other, which is why the visible symptom was a contradiction.

**The rate list was rebuilt from a probe that is allowed to fail.** In `RefreshSampleRateOptions`:

```csharp
int[] availableRates = supportedRates.Count > 0
    ? supportedRates.ToArray()
    : [fallbackSampleRate];        // 44100, hard-coded

int selectedSampleRate = availableRates.Contains(preferredSampleRate)
    ? preferredSampleRate
    : availableRates[0];
```

An empty probe became a list of one constant, the configured rate was not in it, and the selection fell to 44100 — silently, with the next Apply persisting it. Your own comment in `GetSupportedSampleRates` already names the cause: some ASIO drivers refuse a second open moments after the first. Closing the settings window and reopening it is exactly that.

**The status line was written before the selection existed.** `RefreshAsioDriverInfo` writes it at the end of the probe, which runs *before* `RefreshSampleRateOptions` fills the combo — and at that moment `GetSelectedSampleRate()` answers with its own 44100 fallback, because nothing is selected yet. Filling the combo afterwards changed a number nobody re-read. That is how `96000` in the list came to sit above `44100 Hz supported` in green, and it survived fixing the fallback itself: the first build I tested still showed it.

## What it does now

**An empty probe is treated as the absence of an answer, not as an answer.** The list and the user's selection stand, and the line says the driver did not report rather than pronouncing on a rate nobody tested.

**A real fallback is shown as one.** When the driver *does* answer and does not offer the configured rate, the rate still changes — but the line names the one that was taken away: `96000 Hz is not offered by this driver — changed to 44100 Hz`, in salmon.

**The line is written after the selection settles**, so its number and its verdict can no longer come from different moments.

The three cases are decided in `SampleRateOptions.Resolve` rather than inside the panel, so they are testable without a window — `SampleRateOptionsTests` covers a failed probe changing nothing, a driver that offers the rate keeping it, a driver that does not saying so, and the first population (nothing in the combo yet) resolving to the configured rate rather than to the constant. The constant is now reserved for having nothing at all to go on.

## Verification

Same machine, same driver, same repro: ASIO → Focusrite USB ASIO → 96000 → Apply → close the settings window → reopen.

- **Before:** the combo came back at 44100 with `44100 Hz supported` under it, and Apply would have persisted it.
- **After:** the combo comes back at **96000**, and the line under it reads **96000 Hz supported**.

## One thing I did not touch

While verifying, the **Playback latency** line behaves oddly on this setup: it does not change when the sample rate is changed in the combo, but shows a different value after the window is closed and reopened (509 samples at one opening, 419 at another). There is no caching in `GetDriverInfo` — each call opens the driver afresh — so my guess is that NAudio's `PlaybackLatency` reports the device's *current* buffer rather than one implied by the rate being probed, and the probe asks whether a rate is supported without putting the device on it.

That is a guess about a driver, not about your code, and it is outside what this PR claims to fix. Flagging it rather than changing anything blind — you will know immediately whether it is expected.

### DIFF
--- a/source/Options/MeasurementOptions.cs
+++ b/source/Options/MeasurementOptions.cs
@@ -2123,6 +2123,17 @@ namespace Resonalyze.Options
             {
                 RefreshSweepBandPreview();
             }
+
+            // The status line pairs a number taken from the selection with a verdict
+            // taken from the probe, so it can only be written once the selection has
+            // settled. RefreshAsioDriverInfo writes it before this method has filled the
+            // combo, when GetSelectedSampleRate still answers with its own fallback —
+            // which is how "96000" in the list came to sit above "44100 Hz supported"
+            // in green. Writing it again here is what keeps the two halves in step.
+            if (comboBoxAudioBackend.SelectedIndex == (int)AudioBackend.Asio)
+            {
+                UpdateAsioStatusLabels();
+            }
         }
 
         private IReadOnlyList<int> GetSupportedSampleRates()
--- a/source/Options/MeasurementOptions.cs
+++ b/source/Options/MeasurementOptions.cs
@@ -43,6 +43,10 @@ namespace Resonalyze.Options
         private IReadOnlyList<AudioEndpointDescriptor> wasapiRenderEndpoints = Array.Empty<AudioEndpointDescriptor>();
         private IReadOnlyList<AsioDeviceInfo> asioDrivers = Array.Empty<AsioDeviceInfo>();
         private AsioDriverInfo asioDriverInfo = AsioDeviceCatalog.EmptyDriverInfo;
+        // What the last rate probe amounted to, so the status line can say which of the
+        // three situations it is in rather than pronouncing on a rate it did not test.
+        private bool sampleRateProbeFailed;
+        private int? sampleRateFellBackFrom;
         private bool initializing;
         private string? microphoneCalibration0DegreesPath;
         private List<MicrophoneCalibrationDefinition> additionalMicrophoneCalibrations = [];
@@ -1499,12 +1503,30 @@ namespace Resonalyze.Options
             }
 
             int sampleRate = GetSelectedSampleRate();
-            labelAsioSampleRateStatus.Text = asioDriverInfo.SupportsSampleRate
-                ? $"{sampleRate} Hz supported"
-                : $"{sampleRate} Hz not supported";
-            labelAsioSampleRateStatus.ForeColor = asioDriverInfo.SupportsSampleRate
-                ? Color.LightGreen
-                : Color.LightSalmon;
+            if (sampleRateProbeFailed)
+            {
+                // The number is the live selection and the verdict came from the last
+                // probe; when that probe told us nothing the two must not be combined
+                // into a confident sentence about a rate nobody tested.
+                labelAsioSampleRateStatus.Text =
+                    $"{sampleRate} Hz kept — the driver did not report its rates";
+                labelAsioSampleRateStatus.ForeColor = Color.Khaki;
+            }
+            else if (sampleRateFellBackFrom is int previous)
+            {
+                labelAsioSampleRateStatus.Text =
+                    $"{previous} Hz is not offered by this driver — changed to {sampleRate} Hz";
+                labelAsioSampleRateStatus.ForeColor = Color.LightSalmon;
+            }
+            else
+            {
+                labelAsioSampleRateStatus.Text = asioDriverInfo.SupportsSampleRate
+                    ? $"{sampleRate} Hz supported"
+                    : $"{sampleRate} Hz not supported";
+                labelAsioSampleRateStatus.ForeColor = asioDriverInfo.SupportsSampleRate
+                    ? Color.LightGreen
+                    : Color.LightSalmon;
+            }
             labelAsioPlaybackLatencyValue.Text =
                 asioDriverInfo.PlaybackLatency > 0
                     ? $"{asioDriverInfo.PlaybackLatency} samples"
@@ -2063,15 +2085,23 @@ namespace Resonalyze.Options
 
         private void RefreshSampleRateOptions(int preferredSampleRate)
         {
-            int fallbackSampleRate = 44_100;
-            IReadOnlyList<int> supportedRates = GetSupportedSampleRates();
-            int[] availableRates = supportedRates.Count > 0
-                ? supportedRates.ToArray()
-                : [fallbackSampleRate];
+            SampleRateResolution resolution = SampleRateOptions.Resolve(
+                GetSupportedSampleRates(),
+                preferredSampleRate,
+                comboBoxSampleRate.Items.Count > 0);
+            sampleRateProbeFailed = resolution.ProbeFailed;
+            sampleRateFellBackFrom = resolution.FellBackFrom;
+            if (resolution.ProbeFailed)
+            {
+                // Nothing is rebuilt on the absence of an answer: the list and the
+                // user's selection stand, and the status line says the driver did not
+                // report. Rebuilding here is what used to replace a working 96 kHz with
+                // 44.1 and persist it on the next Apply.
+                return;
+            }
 
-            int selectedSampleRate = availableRates.Contains(preferredSampleRate)
-                ? preferredSampleRate
-                : availableRates[0];
+            int[] availableRates = resolution.Rates;
+            int selectedSampleRate = resolution.Selected;
 
             bool wasInitializing = initializing;
             initializing = true;
--- a/source/Options/MeasurementOptions.cs
+++ b/source/Options/MeasurementOptions.cs
@@ -1255,6 +1255,14 @@ namespace Resonalyze.Options
                 return;
             }
 
+            // The user picked this rate, so nothing was taken away from them: the
+            // marker left by an earlier automatic fallback describes an action that is
+            // now over, and UpdateAsioStatusLabels would otherwise keep reporting
+            // "96000 Hz is not offered — changed to 48000 Hz" about a rate the user
+            // chose. The probe verdict is left alone; it is still the last thing the
+            // driver actually said.
+            sampleRateFellBackFrom = null;
+
             // The achieved band and its cycle-quantized duration depend on the
             // sample rate, so re-preview on any change (not just for ASIO). The
             // rate itself belongs to the audio backend group and is not applied
@@ -1897,7 +1905,16 @@ namespace Resonalyze.Options
         private void ValidateSelectedWaveSampleRate(int sampleRate)
         {
             IReadOnlyList<int> supportedRates = GetSupportedSampleRates();
-            if (supportedRates.Count > 0 && !supportedRates.Contains(sampleRate))
+            // Empty is an answer here, never silence: a Wave pair reports no rate in
+            // common only when there is none. Skipping validation on it let the rate
+            // sitting in the combo through to a device that cannot open it.
+            if (supportedRates.Count == 0)
+            {
+                throw new InvalidOperationException(
+                    "Wave devices report no sample rate in common for the current " +
+                    "configuration. Change the devices, the channel counts or the bit depth.");
+            }
+            if (!supportedRates.Contains(sampleRate))
             {
                 throw new InvalidOperationException(
                     $"Wave devices do not support {sampleRate} Hz for the current configuration.");
@@ -2092,7 +2109,7 @@ namespace Resonalyze.Options
                 IsAsioSampleRateProbeFailure());
             sampleRateProbeFailed = resolution.ProbeFailed;
             sampleRateFellBackFrom = resolution.FellBackFrom;
-            if (resolution.Rates.Length == 0)
+            if (resolution.Rates is null)
             {
                 // Nothing is rebuilt on the absence of an answer: the list and the
                 // user's selection stand, and the status line says the driver did not
@@ -2113,6 +2130,9 @@ namespace Resonalyze.Options
 
             int[] availableRates = resolution.Rates;
             int selectedSampleRate = resolution.Selected;
+            // An empty list is a real outcome, not a missing one: no rate works for this
+            // configuration, so the combo offers nothing and Apply refuses. Filling in the
+            // configured rate here is what used to hand the user a rate no device reported.
 
             bool wasInitializing = initializing;
             initializing = true;
--- a/source/Options/MeasurementOptions.cs
+++ b/source/Options/MeasurementOptions.cs
@@ -2088,15 +2088,26 @@ namespace Resonalyze.Options
             SampleRateResolution resolution = SampleRateOptions.Resolve(
                 GetSupportedSampleRates(),
                 preferredSampleRate,
-                comboBoxSampleRate.Items.Count > 0);
+                comboBoxSampleRate.Items.Count > 0,
+                IsAsioSampleRateProbeFailure());
             sampleRateProbeFailed = resolution.ProbeFailed;
             sampleRateFellBackFrom = resolution.FellBackFrom;
-            if (resolution.ProbeFailed)
+            if (resolution.Rates.Length == 0)
             {
                 // Nothing is rebuilt on the absence of an answer: the list and the
                 // user's selection stand, and the status line says the driver did not
                 // report. Rebuilding here is what used to replace a working 96 kHz with
                 // 44.1 and persist it on the next Apply.
+                //
+                // The flags above have just changed, and the last thing to write the
+                // status line was RefreshAsioDriverInfo, before Resolve ran — so it is
+                // still rendered from the previous state and would keep a stale
+                // supported/not-supported sentence. Write it here too, exactly as the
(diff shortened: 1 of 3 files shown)


source/Options/SampleRateOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+    public const int FallbackSampleRate = 44_100;
+
+    public static SampleRateResolution Resolve(
+        IReadOnlyList<int> supportedRates,
+        int preferredSampleRate,
+        bool hasExistingList)
+    {
+        ArgumentNullException.ThrowIfNull(supportedRates);
+
+        if (supportedRates.Count == 0 && hasExistingList)
```
DIMOSUS: [P1] Please don't infer “probe failed” from an empty rate list here. `Resolve` is called for every backend, but `GetSupportedSampleRates()` deliberately returns `[]` for non-ASIO cases such as WASAPI Shared endpoints whose mix rates differ, and Exclusive/Wave configurations with no common candidate. With an existing combo this now preserves the previous device/backend's rate list; `ValidateSelectedWaveSampleRate` also treats an empty support list as non-blocking, so an old rate can be offered and persisted for the new configuration. The keep-current-list behavior needs an explicit probe-failure signal scoped to the ASIO failure case, rather than `Count == 0`.

source/Options/MeasurementOptions.cs
```diff
@@ -2063,15 +2085,23 @@ private void HandleAudioConfigurationChanged()
...
+                comboBoxSampleRate.Items.Count > 0);
+            sampleRateProbeFailed = resolution.ProbeFailed;
+            sampleRateFellBackFrom = resolution.FellBackFrom;
+            if (resolution.ProbeFailed)
+            {
+                // Nothing is rebuilt on the absence of an answer: the list and the
+                // user's selection stand, and the status line says the driver did not
+                // report. Rebuilding here is what used to replace a working 96 kHz with
+                // 44.1 and persist it on the next Apply.
+                return;
```
DIMOSUS, line 2142: [P2] This early return skips the only status refresh that happens after `sampleRateProbeFailed` / `sampleRateFellBackFrom` are updated. `RefreshAsioDriverInfo()` already called `UpdateAsioStatusLabels()` before `Resolve`, so this path leaves the label rendered from the previous state; the new “driver did not report its rates” branch cannot be reflected here and a stale supported/not-supported sentence can remain. Please refresh the ASIO status before returning, or move status rendering after resolution for all paths.

DIMOSUS (changes requested):
I found two regressions in the new sample-rate resolution/state handling that should be fixed before merge.

ayukhno:
Both addressed in `a91b255`.

## P1 — the probe failure is told now, not inferred

You were right that `Count == 0` cannot carry this. `Resolve` runs for every backend, and an empty list is a
real answer almost everywhere — WASAPI Shared endpoints whose mix rates differ, an Exclusive or Wave pair with
no rate in common. Preserving the previous device's list there was wrong on its own, and with
`ValidateSelectedWaveSampleRate` also treating an empty support list as non-blocking, a rate belonging to
another configuration could be offered and then persisted.

`Resolve` now takes `probeFailed` from the caller. The decision itself went into `SampleRateOptions` as
`IsProbeFailure(isAsioBackend, asioDriverName, reportedRateCount)` rather than into the panel — for the same
reason this PR put `Resolve` there in the first place: so it can be tested without a window. Leaving it in the
form would have made the new scoping the one part nothing covers.

It is true only for a **named ASIO driver that reported no rate**. A driver that opens accepts at least one
standard rate, so naming none is silence rather than an answer. With no driver selected there is nothing to
preserve — which is the same test `RefreshAsioDriverInfo` already applies when deciding whether the saved
channel routing survives, so the panel now answers this question the same way twice.

One consequence worth stating rather than leaving to be discovered: a failed probe with **nothing on screen to
keep** now reports `ProbeFailed` anyway. It still builds a list — the configured rate standing alone — but that
rate is not something a driver offered, so the status line must not call it supported.

## P2 — the status line

Fixed as you asked: the early return refreshes the ASIO status before returning, so the new "did not report"
branch can actually appear there instead of leaving whatever `RefreshAsioDriverInfo` rendered from the previous
state.

The return is also keyed on there being nothing to rebuild rather than on `ProbeFailed`. That is what lets the
two come apart in the first-population case above — a failed probe that still has a list to build.

## Tests

Seven, including the empty-but-real answer that regressed, the failed probe with and without a list to keep,
and `IsProbeFailure` across ASIO and non-ASIO with a named driver, an empty name and a null one.

## What I have not verified

The hardware behaviour, and I would rather say so than imply otherwise. The original #92 repro — closing and
reopening the settings window so the driver refuses a second open — I cannot reproduce here, and the P2 status
label is a `Form` the App suite never constructs, so neither is covered by a test. Both are reasoned from the
code and both compile; neither has been watched happening. If you would rather see that before merging, say so
and I will run it on Windows.

source/Options/SampleRateOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(supportedRates);
+
+        if (probeFailed && hasExistingList)
+        {
+            return new SampleRateResolution([], preferredSampleRate, true, null);
+        }
+
+        int[] rates = supportedRates.Count > 0
+            ? [.. supportedRates]
```
DIMOSUS: [P1] A real empty answer is still turned into an unsupported synthetic option here. For non-ASIO backends `probeFailed` is false, and `supportedRates == []` means there is no compatible rate for the current configuration. This branch replaces that with `[preferredSampleRate]`, so the combo offers a rate that was never reported as supported. For Wave, Apply still does not catch it because `ValidateSelectedWaveSampleRate` skips validation when the support list is empty. Please preserve the distinction: synthesizing the configured rate makes sense for a failed probe with nothing on screen to keep, but a genuine empty answer should not manufacture a supported option and should be rejected on Apply.

source/Options/MeasurementOptions.cs
```diff
@@ -1499,12 +1503,30 @@ private void UpdateAsioStatusLabels()
...
+            if (sampleRateProbeFailed)
+            {
+                // The number is the live selection and the verdict came from the last
+                // probe; when that probe told us nothing the two must not be combined
+                // into a confident sentence about a rate nobody tested.
+                labelAsioSampleRateStatus.Text =
+                    $"{sampleRate} Hz kept — the driver did not report its rates";
+                labelAsioSampleRateStatus.ForeColor = Color.Khaki;
+            }
+            else if (sampleRateFellBackFrom is int previous)
```
DIMOSUS, line 1547: [P2] `sampleRateFellBackFrom` is sticky across a manual rate change. After an automatic 96 kHz → 44.1 kHz fallback, selecting 48 kHz triggers `RefreshAsioDriverInfo()` but never clears/re-resolves this field, so `UpdateAsioStatusLabels()` keeps taking this branch and reports that 96 kHz was “changed to 48000 Hz” even though 48 kHz was explicitly selected by the user. Please clear the fallback marker on a user rate change (or derive this status from the current action rather than persistent form state).

DIMOSUS (changes requested):
The two previous review threads are fixed in their original form. I found two follow-up state/validation problems that still need addressing before merge.

ayukhno:
Both fixed in `a1d4db0`.

## P1 — the empty answer

You are right, and the way you put it is the diagnosis: I scoped the failure signal on the way *into*
`Resolve` and left the way *out* of it unchanged. The synthesising branch still fired on any empty list,
including the one case that is a real answer.

There are three outcomes rather than two, and the record now says which is which:

| `Rates` | meaning |
|---|---|
| `null` | nothing to **rebuild** — the list on screen is the last real answer anyone got |
| empty | nothing to **offer** — there is an answer, and it is that no rate works here |
| non-empty | offer exactly these |

Synthesising the configured rate is now reachable only from a failed probe with no list to keep. A genuine
empty answer produces an empty list, and `ValidateSelectedWaveSampleRate` refuses it by name instead of
skipping validation:

> Wave devices report no sample rate in common for the current configuration. Change the devices, the channel
> counts or the bit depth.

## P2 — the sticky marker

Cleared on a user rate change, in `comboBoxSampleRate_SelectedIndexChanged`, before anything re-renders. The
marker described an automatic action that is over the moment the user picks a rate themselves.

I deliberately did **not** clear `sampleRateProbeFailed` alongside it. That one is not a record of an action,
it is the last thing the driver actually said, and "48000 Hz kept — the driver did not report its rates" stays
true after a manual pick. Say the word if you would rather both were cleared.

## One test changed meaning, not wording

`TheFirstPopulationHasNoListToKeep` asserted that an empty probe with nothing in the combo falls back to the
configured rate. Under your P1 that is no longer what an empty *answer* does — only what a failed *probe* with
nothing to keep does. I split it rather than edit the assertion, so the change is visible in the diff instead
of looking like a test quietly rewritten to match new code. Twelve cases now, including the genuine empty
answer both with and without a list on screen.

## Verified

App 1047/1047, Audio 142/142, DSP 1109/1109 on Windows, plus the publish and installer steps — I run the
branch through my own fork's CI before it reaches you, since the App suite cannot run on macOS.

Still not verified, and I would rather repeat it than let it pass quietly: the hardware behaviour. The #92
repro and the status label are `Form` paths no test covers. Both are reasoned and both compile; neither has
been watched happening.

source/Options/SampleRateOptions.cs
DIMOSUS: The original synthetic-option problem is fixed in `Resolve`, but one Apply path still keeps this issue open: WASAPI Exclusive can legitimately produce an empty common-rate list, which now clears the combo. `GetSelectedSampleRate()` then falls back to 44.1 kHz, and `SetOptions()` has no equivalent of `ValidateSelectedWaveSampleRate` for Exclusive, so Apply can still proceed with a rate that `GetSupportedSampleRates()` just determined is unsupported by the endpoint pair. Please reject an empty/unsupported Exclusive rate on Apply rather than letting the combo fallback become configuration.

source/Options/MeasurementOptions.cs
DIMOSUS: Fixed in `a1d4db0`: the manual sample-rate change clears `sampleRateFellBackFrom` before the ASIO re-probe, so the stale automatic-fallback message no longer survives a user selection.

source/Options/SampleRateOptions.cs
DIMOSUS: Duplicate reply; please ignore this one.

DIMOSUS (changes requested):
One new state bug remains in the manual ASIO sample-rate path: `comboBoxSampleRate_SelectedIndexChanged` clears `sampleRateFellBackFrom`, then immediately runs a fresh `RefreshAsioDriverInfo()`, but `sampleRateProbeFailed` is deliberately left untouched. If the previous probe failed and this new probe succeeds, `asioDriverInfo` now contains a valid fresh result while `sampleRateProbeFailed` is still `true`, so `UpdateAsioStatusLabels()` continues to report “the driver did not report its rates” for a probe that actually succeeded. Please update/re-resolve the probe-failure state from the new `asioDriverInfo` result before rendering status. I also left the existing P1 thread open with the remaining WASAPI Exclusive Apply case.

DIMOSUS (approved):

ayukhno:
Both fixed in `59f9ff5`.

**The stale probe flag — you were right, and my round-two reasoning was wrong about its own handler.**
I argued that `sampleRateProbeFailed` records what the driver said rather than an action, and so should
survive a manual pick. But `comboBoxSampleRate_SelectedIndexChanged` does not only pick a rate: it calls
`RefreshAsioDriverInfo` to re-probe for the latency figures. That probe is a fresh answer, so keeping the
old verdict beside a fresh `asioDriverInfo` is exactly the inconsistency you describe — a probe that has
since succeeded still rendered as "the driver did not report its rates".

The verdict is now settled where the probe happens, immediately after `AsioDeviceCatalog.GetDriverInfo`,
so every caller gets it whether or not it goes on to rebuild the rate list — and a manual rate change is
precisely the caller that does not. `RefreshSampleRateOptions` still takes the same predicate for the
resolution it acts on; it is the same value from the same probe, so the two cannot disagree.

**WASAPI Exclusive on Apply.** Agreed, and the hole was exactly where you put it: since the last round an
Exclusive pair with no rate in common correctly empties the combo, and an empty combo is where
`GetSelectedSampleRate` answers with its own 44.1 kHz — so Apply committed a format the endpoints had just
refused. The combo fallback becoming configuration is the thing this PR exists to stop, so it had no
business surviving one enum value away from Wave.

The decision now lives in `SampleRateOptions.ValidateSelectedRate`, beside `Resolve` and `IsProbeFailure`
and for the same reason — it is testable without a window — and Wave and Exclusive both call it. Two
scoping notes:

- **ASIO stays out of it on purpose.** There an empty list can be silence, and refusing on silence is the
  original bug.
- **Shared needs nothing.** It overwrites `sampleRate` with the endpoints' mix rate and never reads the
  combo, so no fallback of the combo's can reach the configuration through it. I checked rather than
  assumed, because my first instinct was to widen the fix to both WASAPI modes.

The Exclusive check sits after the endpoint-availability test, so a disconnected endpoint keeps its own
message instead of being reported as a rate mismatch.

Three tests on the new validator: the empty list, a rate outside the list, and one inside it.

Verified on Windows before pushing — App 1050/1050, Audio 142/142, DSP 1109/1109, publish and installer
green. The status-label path itself is still `MeasurementOptions`, a Form the App suite never constructs,
so the first fix has no test: only a run on Windows with the interface can close it.

PR #117, opened 2026-08-24
Read a junction's arrival coherence per frequency, not just per lag

DIMOSUS:
## What this adds

A **Coherence** mode beside Correlation on the Virtual DSP junction plot, reading the same junction *per frequency* instead of per lag.

A sub-band probe slides across the pair band (sixth-octave steps, 2/3-octave width, each band's direct-sound cut sized to its own period, so the window travels with the frequency it measures). Per band the envelope of the band-limited GCC-PHAT gives:

- **Δt to optimum** — the delay that would center that band's arrivals, drawn inside the dashed **±T/2** comb corridor;
- **r current** (dotted) against the shaded **r attainable** ceiling — the gap is what the applied tune leaves on the table, and it closes where a band is centered.

Reading it: a flat Δt at zero is a settled junction; a flat line *offset* from zero is a junction that missed by whole periods; a sloped or stepped one is dispersion that no single delay reconciles.

The ladder deliberately reports **no polarity**. A 2/3-octave probe makes a coherence packet `2 / (2^(B/2) − 2^(−B/2))` ≈ **4.3×** wider than the spacing between opposite-signed comb lobes — the ratio is free of frequency — so those lobes sit inside the packet's own plateau and the carrier's sign at the maximum is noise. Measured across all twenty archived junctions the envelope falls 0–6% between a band's optimum and its neighbours, on sharply tuned junctions as much as on rough ones. Polarity is what the **correlation view** answers, over the pair's whole band, where the lobes genuinely separate.

The correlation view also gains each comb's **± analytic envelope** as thin translucent guides in the parent curve's color, kept out of the legend — the coherence packet's center, readable against the carrier lobes it is made of.

## Guards

- Bands where one driver has fallen **25 dB** below the other are dropped: GCC-PHAT ignores level by design and would otherwise read confident "coherence" off a crossover remnant the sum cannot hear.
- Junctions below **120 Hz** carry a note that the long band windows let cabin modes rule the read.
- The mode is a **diagnostic for manual tuning**: Auto delay keeps its own guarded estimators and never reads this plot.
- Persists as an additive flag, so an older build opens such a session on the magnitude view.

## Junction phase read-out, trimmed to what a tuner can act on

`φfc  fix ms  score`, replacing `φfc  fix ms  lobe`:

- the **lobe margin** left for the tooltip — a difference of two phase scores has no scale a reader can judge, and its one decision (trust the fix or not) is already the `!` flag;
- a **fix** worth less than 10° of phase at fc (0.03 dB in the sum) renders `·` instead of a number inviting a correction that changes nothing; above ~5.6 kHz such a fix is smaller than 0.005 ms, so it takes a third decimal rather than rounding to "0.00";
- **score** is new: the band's phase-alignment score as the junction STANDS (−1…+1), the quantity the fix maximizes. It reads without a legend and moves while a delay is dragged. On the archive it separates junctions the old block rendered alike — the owner's re-tuned Passat v2 reads 0.89 / 0.93, while v2's 220 Hz junction reads 0.01 beside a −5.68 ms inverted fix.

## Performance

Both junction modes were too slow to tune against. Measured on the archived cabins at 96 kHz:

| | before | after |
|---|---|---|
| coherence ladder | seconds | **140–302 ms** |
| correlation view | 930–1250 ms | **167–372 ms** |

The ladder finds each channel's front once per channel (not per band) and correlates direct-sound cuts **trimmed to the windows' own span** — both channels sliced by ONE shared offset, so every relative position, and therefore every lag, is unchanged while the FFTs shrink from the record's length to the window's; bands run in parallel. The correlation view builds its alignment bins **once for both score polarities** (a bit-exact factoring, pinned by a test), probes them in parallel, trims the direct twin the same way, and computes its four independent reads side by side. The full-record PHAT deliberately stays untrimmed — the whole capture is that curve's subject.

Trimming caught its own trap on the way in: a short buffer's circular correlation aliases its lobes into the far lags (+2.75 ms on a synthetic pure delay). The buffers are padded to the searched lag range, and the pure-delay test is what fails without it.

## Fixes found along the way

- **Crash on opening a session over a loaded one** (older than this branch): importing rebinds every channel's runtime state on the UI thread while renders are in flight, so a consumer reading the sample rate off the **live** channel saw a zero against a real response. `ProcessedChannel` now carries the rate it was processed at.
- **Silent cancellation**: a stale render is reported by returning `null` instead of throwing out of a parallel body, which a Just My Code debugger stopped on at every delay edit (25 exceptions per edit burst, now zero). External cancellation still propagates — including when it arrives mid-computation, which the entry guard used to hide.
- **Coherence frequency axis** re-fits when the junction's band changes; the invalidation state records the band, not just the pair and the lag bucket.

## Verification

- CI on the branch: **2650 passed, 1 skipped, 0 failed** (142 Audio + 1203 DSP + 1305 App), plus publish and installer.
- New tests: ladder behaviour (pure delay → flat, inverted pair stays centered, level gate, dispersive junction), `BuildCoherenceView` in the correlation view's frame, one-kind-of-marker with no polarity series, axis re-fit on a band change, both-polarity sweep equality, envelope guides outside the legend, coherence-mode persistence round-trip, snapshot-rate regression, and two mid-flight cancellation contracts.
- Every behavioural fix was falsified before shipping: the pin was run against the pre-fix code and observed to fail.
- The archived cabins were re-rendered through the **product** code path (`DrawCoherence` / `DrawCorrelation`) after every change, and the reports diffed.

README documents the mode, the polarity caveat and the read-out in the Virtual DSP section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1563,7 +1563,7 @@ delay on the lower channel that would maximize the overlap-band phase score,
 with `i` recommending a flip, `~` warning that a flip nearly ties, and `!`
 warning that the overlap band is too narrow to rule out a whole-period hop — so
 the fix is not to be trusted, and the junction's coherence ladder is the better
-read. A fix worth less than 10° of phase at the crossover (0.13 dB in the sum)
+read. A fix worth less than 10° of phase at the crossover (0.03 dB in the sum)
 shows as `·`: a settled tune has nothing to apply there, and a number would
 invite a correction that changes nothing. Last comes **score**, where the
 junction stands as it is — the band's phase-alignment score (−1…+1) that the
--- a/README.md
+++ b/README.md
@@ -1565,8 +1565,12 @@ warning that the overlap band is too narrow to rule out a whole-period hop — s
 the fix is not to be trusted, and the junction's coherence ladder is the better
 read. A fix worth less than 10° of phase at the crossover (0.13 dB in the sum)
 shows as `·`: a settled tune has nothing to apply there, and a number would
-invite a correction that changes nothing. The tooltip carries every fitted
-figure behind those two columns, including the lobe margin the `!` is drawn
+invite a correction that changes nothing. Last comes **score**, where the
+junction stands as it is — the band's phase-alignment score (−1…+1) that the
+fix maximizes: 1.00 is aligned across the overlap, 0 a wash, negative means the
+two drivers are subtracting. It moves while a delay is dragged, so it answers
+"is this getting better", which the fix alone cannot. The tooltip carries every
+fitted figure behind those columns, including the lobe margin the `!` is drawn
 from. A **Δ L−R** block
 below reports each pair's inter-side state — the two sides' band-limited envelope
 arrivals with their difference (positive means the right side leads, the scene
--- a/README.md
+++ b/README.md
@@ -1528,6 +1528,30 @@ inverted rival, the wavefronts within a period or two of the front decide it,
 because that is the part of the record the drivers made and the room had not yet
 answered.
 
+The plot's **Coherence** mode reads the same junction per frequency instead of
+per lag: a sub-band probe slides across the pair band (sixth-octave steps,
+2/3-octave width, each band's direct-sound cut sized to its own period), and
+per band the envelope of the band-limited GCC-PHAT gives **Δt to optimum** —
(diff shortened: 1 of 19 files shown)


DIMOSUS:
@codex review

PR #126, opened 2026-08-26
Design the filters at the DSP's rate, not the microphone's

DIMOSUS:
Virtual DSP simulated every filter at the rate the channels happened to be **measured** at, which is only right when the sound card and the processor agree. The bilinear transform warps a corner by the rate it was designed for, so a 48 kHz card could not design for a 96 kHz processor — and nothing in the code checked or even mentioned it.

## The split

`PreparedDspResponse` builds its cascade at the **processor's** rate; `ApplyToSpectrum` takes the **record's** rate for the bin grid. Bin `i` of an N-point FFT lands at `ω = 2π·i·fs_signal/(N·fs_processor)` on the processor's unit circle, so a 48 kHz measurement reads a 96 kHz chain over the lower half of that circle — exactly the band a 48 kHz record can carry. This is exact, not approximate: a chain is LTI and invents no frequency its input lacks.

Three details follow from it:

- bins past the processor's own Nyquist are zeroed (a device reconstructs nothing up there, and this is what makes the same setup measured at 96 and 192 kHz simulate alike);
- the bulk delay rides the record's grid — it is a time, not a sample count;
- the filter tail is sized in milliseconds before it is counted in samples.

## The processor a project is designed for

`DspProcessorCatalog` is one line per device — 41 of them across AMP, HELIX, Audison, Hertz, Mosconi, ESX, miniDSP and JL Audio — with the id derived from the names, so adding a device is a single line and nothing else. The profile (model, processing rate, Q convention) is stored in the project file and edited through a new **DSP processor** dialog, which sits above Auto crossover / Auto delay. Pick a model and its properties are locked to the catalog; pick Custom and state them yourself.

The profile travels into the EQ Wizard with a PEQ handoff, where it now decides the rate the bank is fitted and exported at, and the Q convention the tuning sheets are stated in. The wizard's own two selectors show it locked for as long as that source is loaded, and the user's own choice is kept for everything else.

A project file from before this opens as **Custom following its measurements**, so no existing tune changes on its own.

## What it measures

Real cabin (v6), right tweeter — HP 1800 Hz BW48 plus 20 PEQ bands — against the 96 kHz record processed at 96 kHz:

| variant | shape rms | worst | waveform residual |
| --- | --- | --- | --- |
| 48 kHz resample, processor 96 kHz | 0.000 dB | 0.001 dB | −84 dB below peak |
| 48 kHz resample, processor 48 kHz | 0.960 dB | +4.08 dB at 19.9 kHz | −5.9 dB |
| 96 kHz record, processor 48 kHz | 0.960 dB | +4.08 dB at 19.9 kHz | — |

The last two agreeing is the whole point: the error belongs to the design rate, not to the record. By stage, the crossover contributes 0.049 dB rms and the PEQ bank 0.949 — the bands sit at 10–20 kHz, where a 24 kHz Nyquist is close.

Auto delay agrees. With the processor set correctly, 48 kHz and 44.1 kHz resamples of all seven measurements reproduce the 96 kHz proposal across both sides, to the 0.01 ms grid the engine rounds onto (44.1 kHz moves three channels of eight by one step, 10 µs — under half a sample). With the processor set wrong, the right tweeter jumps a whole period: 9.44 ms against 8.73, because the PHAT extremum at that junction stops being edge-pinned and reseeds the search window past the right answer.

## Tests

`ProcessorSampleRateTests` pins the equivalence (a low-rate record carries a high-rate chain), the numbers a wrong design rate produces, that the delay is a time, that the tail is sized in time, and the Nyquist gate. `DspProcessorCatalogTests` guards the derived ids against collision and pins each device's properties. The project file round-trips the profile, and a file without one still follows its measurements. Full suite green: 2919 passed.


DIMOSUS:
Both Codex findings were real; fixed in cfe6df8.

**Handoff guard.** The return token now carries the processor's rate and `TryApplyReturn` refuses when the project has moved to another one — a bank fitted for a 96 kHz device can no longer be installed into a project running at 48 kHz. Only the rate is guarded: the profile's Q convention moves printed numbers and leaves the tune exactly as fitted. Pinned by `ReturnAfterTheProjectChangedProcessors_Refuses` (and `AHandoffRecordsTheProcessorItWasTakenUnder`).

**Nyquist gate on the fast path.** Right: a bypassed or gain-only channel kept the band above the processor's Nyquist while any filtered channel lost it, so the members of a sum were band-limited differently depending on whether a channel happened to carry a filter. `ApplyChain` now takes the FFT-free path only when the record has nothing above the processing rate to lose (`PreparedDspResponse.CanScaleInTimeDomain`); at or below that rate it stays one multiply per sample. Pinned by `AScaleOnlyChainIsBandLimitedToo_WhenTheRecordOutrunsTheProcessor`, which also checks the cheap path is still cheap.

Full suite green: 2922 passed.

DIMOSUS:
Thanks — all three addressed.

**1 and 2 were already fixed in cfe6df8**, which the review predates (it read `3db6a1a`). The token now carries the processor rate and `TryApplyReturn` refuses when the project moved (`ReturnAfterTheProjectChangedProcessors_Refuses`), and the FFT-free path is taken only when the record has nothing above the processing rate to lose — `CanScaleInTimeDomain` — with a test that runs Identity AND gain-only at 192 → 96 kHz, exactly the case the LowPass test could not reach.

**3 was right, and it is fixed in 37d35f8.** Storing `null` whenever the number happened to match the measurements made two different intentions indistinguishable, and it made the choice unconfirmable on a legacy project: re-opening the dialog and pressing OK compared equal and returned early. The rate list now offers **Follow measurements (48 kHz)** beside the fixed rates, so the intent is what the user picks and what the file stores; stating 48 kHz keeps 48 kHz when the measurements are replaced at 96. The rules moved onto the project file — `ResolveDspProcessor`, `SetDspProcessor`, `DspProcessorRateFollowsMeasurements` — so they are testable without a form, and the panel compares the intent as well as the numbers before deciding a dialog changed nothing. Three tests pin it: a stated rate survives the measurements being replaced, a following one tracks them, and a named model never follows.

**On catalog provenance** — agreed, and the file now says so: the table is the owner's, read off the makers' published processing rates, with only Panacea's Q convention confirmed by measurement. The tests pin what the file says, which is not the same as pinning how a device behaves; a correction to a line is a data fix and reaches every project naming that model, since a named model always resolves from the catalog.

Full suite green: 2925 passed.

DIMOSUS:
Right on both counts — fixed in e26768d.

**The Custom → preset → Custom regression was real, and it was my mistake in a plain way:** the line that carries `customFollowsMeasurements` was written, but the script that wrote it aborted partway and never saved the file, so the previous commit shipped without it. Your reproduction is exactly what happened.

`DspProcessorDialogTests` now drives the real dialog on an STA thread through both directions of that round trip — follow → state 96 kHz → preset → Custom, and stated → follow → preset → Custom. I checked they earn their keep: with the fix removed, both fail and the other two still pass.

**The `measurementSampleRateHz == 0` mismatch was real too.** The panel was handing over `ProjectSampleRateHz`, whose whole job is to substitute 48 kHz when there is no measurement, so an empty project read "Follow measurements (48 kHz)" as if it held one. It now gets what the project really has — `MeasuredSampleRateHz`, zero included — and the entry says only "Follow measurements" until there is a rate to name. Pinned by `WithoutAMeasurement_FollowingStillResolvesToAUsableRate`.

Full suite green: 2929 passed.

PR #133, opened 2026-08-28
Name the array's controls as they read, and put two figures under them

DIMOSUS:
The manual described the microphone array from the pull request that built it rather
than from the panel that shipped it, and the difference is the kind a reader cannot
recover from: it sent them looking for a button that does not exist.

**The controls now read as they read.** There is no **Array...** button — the row is
**Array microphones** and its button is the count, saying *None...* until one is
configured; the field the docs called a "note" is labelled **Position**. Both names are
corrected in `MANUAL.md` and in the two places `REFERENCE.md` repeated them.

**The positions are counted correctly.** "One row per position" is off by one: the
measurement microphone IS a position — the anchor every other is levelled onto — so the
seven positions the feature was developed against are six rows plus it, and one row is
the least an average can be built from. A reader following the old sentence would have
bought a microphone nobody needs.

**Three things a tuner has to know were missing.** An array wants **ASIO** in practice
(MME reaches two channels in all, and a WASAPI endpoint usually presents an interface's
inputs as stereo pairs). It is bound to the **device** it was configured on, so after
selecting another capture device nothing is recorded until the list is confirmed with
OK. And a capsule's calibration can only be picked from Record Settings' own list, which
means adding it there first — **More calibrations → Manage... → Add file...** — a step
the array dialog cannot do for you and never mentioned.

**The promise in Section 4 had nowhere to land.** It says the positions' disagreement is
stored beside the average "so you can see where it is a claim and where it is not", and
nothing in the guide said where to look. Section 5 gains **Check the array, if you
measured with one**: the three curves (**Show array average**, **Show array
microphones**, **Show array spread**), that they are read from the file and so do not
follow the impulse gate, and how to read the spread — 11 to 12 dB apart is ordinary,
past 20 dB the average is carried by whichever position was loudest and Auto Tune places
no boost there. Section 3 gains the array in the list of what the rig may include, and
Section 10 notes that a still-mounted array answers the spatial-robustness check by
itself. The version banner moves to **v0.7.3**, which carries both the array and the DSP
processor selector the old banner promised for a future release.

## Two figures, and the scene that takes them

`manual/array-microphones.png` and `manual/array-curves.png` are new, and the screenshot
tool takes both, from one measurement named by a new optional `arrayMeasurement` key.
Material a config does not have is skipped with a note in a SWEEP — not every rig has an
array to shoot — but a shot asked for **by name** fails instead, because a named figure
that was never written must not exit 0.

The dialog's rows are read back out of the same measurement the curves are drawn from,
so the two figures describe one set. It is constructed rather than reached through Record
Settings, which offers only inputs the ATTACHED interface has — without the rig that shot
would be an empty dialog. What a measurement does NOT record, the tool no longer invents:
the device's input count, the loopback's input and the backend are declared in
`arrayRig`, and without them the dialog figure is not taken at all. It is the figure's
rig rather than the measurement's — a seven-position set may have been assembled on a
two-microphone interface — so which device the dialog is shown for is stated by the
author instead of composed by the tool, and a rig contradicting the measurement's own
channels stops the shot. The parenthesis it prints comes from the application's own
wording, now shared by the panel and the tool through `ArrayInputSources`. The same key
names each capsule's calibration: the set behind the figure was recorded by one
microphone moved between sittings, so the file stores one name seven times, and a column
that happens to repeat reads in a figure exactly like a column that cannot vary — while
a real array is individually calibrated, as the guide says two sections later.

The curves shot switches distortion, noise floor and coherence OFF: six more traces over
the same decade are what bury the four the figure is about.

The caption's numbers are measured off the file it shows: below 500 Hz the positions sit
2 to 5 dB apart and the point response IS the average to within half a decibel, while
above it they part by some 13 dB and the point response leaves the average by 3 dB in
the median and 19 dB at worst.

## The EQ Wizard figures, and one defect behind them

`eq_wizard`, `eq_wizard_phase`, `manual/eq-wizard-handoff` and `manual/eq-wizard-tuned`
are re-shot: the panel gained an **EQ curve** checkbox and a band guide in #132, and the
committed figures predate both. The session they come from has moved on since the last
sweep, so the text moved with it — the handoff now takes the RIGHT side (**Ch C · R**)
and Auto Tune spends nine bands for 3.7 → 1.5 dB rms, where the captions said left and
"eight bands, 3.9 → 1.1".

`REFERENCE.md`'s caption for the phase figure was stale before this run and had been
since the tool first shot it: it described a midbass zoomed onto a 350 Hz junction with
an all-pass in slot 4, and the figure is a midrange over the full band with no all-pass
at all. Rewritten to the figure that is actually there.

**The one code change is a receipt that lied.** The wizard's header names what it is
about to equalize, and it hard-coded `MMM` for any spatial average — so a channel
measured with an array arrived as **Ch C · R (DSP, MMM)** while the card it came from
said **Array**, sending its reader to look for a capture that does not exist in an array
project. It now reads the document's own `Method`, the way the channel button does.
Tested, and the test was checked against the unfixed code first.

## Figures still owed

`assets/images/manual/record-settings.png` is a committed artifact — two windows
composed by hand with numbered callouts — and it still shows the *Confirm each run*
checkbox removed in #132 and no **Array microphones** row. The tool cannot take it and
it needs the interface attached; the prose beside it already describes the current panel.

Tests: the full suite, 1689 App and 1338 Dsp — the array receipt, and the input-source
wording the figure now shares with the panel.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


DIMOSUS:
Both review points fixed, plus one the figure earned on its own.

**[P2] A named shot that cannot run now fails.** `Scene.Unavailable` still skips in a sweep — a rig without an array should not break the run — but when shots were named on the command line it exits 1 with the reason. Every scene reached with a non-empty request list is one the user asked for by name, so that is exactly the discriminator. Verified: `-- manual/array-microphones` against a config with no `arrayMeasurement` exits 1, and the same without `arrayRig` exits 1.

**[P3] The dialog no longer invents a device.** The input count, the loopback's input and the backend are declared in the config's `arrayRig`, the figure is not taken without them, and a rig contradicting the measurement's own channels stops the shot (verified: loopback on an input a position occupies). The measurement format is untouched, as you suggested. Two notes on the framing: it is the FIGURE's rig rather than the measurement's — the owner's interface takes two microphones, so the seven-position set was assembled, and which device the dialog is shown for is now an authored decision instead of a silent one — and the status line's parenthesis is no longer worded by the tool at all: `ArrayInputSources.Describe` holds it for the panel and the tool together, tested including the stereo-WASAPI branch that tells a user their further inputs need ASIO.

**Also:** every row showed the same `0deg`, for the same reason the rig had to be invented — one microphone moved between sittings stores one calibration name seven times. `arrayRig.calibrations` names them per row, so the figure shows the individually calibrated array the guide describes.

The array scene splits in two, since only the dialog needs the rig. Suite: 1689 App, 1338 Dsp.

PR #134, opened 2026-08-28
Re-take the Record Settings figures on the current panel

DIMOSUS:
Both figures still showed the panel of two releases ago: the *Confirm each run*
checkbox that #132 removed, and no **Array microphones** row, which #130 added and
the manual has described since yesterday. A reader following section 4 would have
looked for a row the picture does not have, next to a checkbox that no longer
exists.

The panel is re-shot on a configured rig — 96 kHz, the array row reading *None...*
— and lands in both places it is used: `measurement-options` in the reference, and
the manual's composition beside the shell's transport column, whose six numbered
callouts now sit on the panel's own groups (the calibrations box gained the array
row; nothing else moved). The Sample Rate row keeps its cyan outline, which is what
section 4's "(outlined)" points at.

The tool's README says why this one is still hand-taken: it runs in portable mode
with its own empty settings, so it would photograph a panel of defaults instead of
a rig, which is the same reason `measurement-options` is asked for by name with the
interface attached.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


DIMOSUS:
Fixed on `main` in d90c833 — the finding was right. Checked the previous figure before changing anything: it carried a three-pixel amber drop at x=122–124 from the box's bottom edge down to the badge, which is why badge and outline were one connected shape there. The recomposition dropped it. The other five badges stand in the gutter and reach their groups horizontally; the sixth stands under the transport column, where there is no gutter, so it is the one that needs a vertical leader — restored to the same 3 px.

PR #135, opened 2026-08-28
Give the microphone calibration one home, and read every measurement through it

DIMOSUS:
The microphone a measurement was taken through was decided in a VIEW. That one fact
produced every defect below, and the manual walked readers straight into the worst of
them: *"select the 90° microphone calibration under the main graph"* — after the
sweeps, which is the one moment at which it labels nothing.

## The calibration gets one home

**Record Settings → Measure through** names the curve this rig records with, beside
the array microphones whose calibrations were already chosen there. A run freezes it
into its result; Save writes that frozen curve into the file. Nothing about a chart
decides it any more, and the confirmation dialog that used to ask about a calibration
adopted from a foreign file is gone with its two helpers — seventy-one lines that
existed because the decision was in the wrong place.

Settings schema **12** takes the rig's selection from the view that used to stamp the
files, so the next sweep is labelled exactly as the last one was, and drops the copies
that could drift: Phase and Group Delay stored ids nothing applied (they read timing,
not level), and Live Spectrum kept a second answer to a question the rig already
answers.

## Every measurement is read through the microphone that took it

**Own (as measured)** is an entry in the analysis calibration list, and every new
measurement selects it — a finished sweep, an opened Resonalyze file, a REW export, a
recorded WAV, a history entry stepped back to. It names a RULE rather than a curve, so
it persists and means the same thing against whatever is open next session, and it is
marked unavailable while the open measurement carries no calibration: "corrected by
nothing" and "corrected by the microphone that measured it" must not look alike.

That retires `FileCalibrationSelection` — the "(from file)" pseudo-entry, the
displaced-selection bookkeeping, the save-time strip for an id that could not be
stored — and closes the case none of it covered: a sweep taken after a foreign file
left the view naming the stranger's calibration while the plot had already moved on.

## A live capture belongs to the run that took it

The blocker. `BuildLiveCaptureDocument` re-renders the accumulated bins when a capture
is SAVED, and it read the calibration live. So: walk an MMM pass through A, change
*Measure through* to B before saving — an edit that does not restart the analyzer when
the audio session is unchanged — and the stored curve was recomputed through B while
the file claimed to have been taken with it. Silent corruption, and the saved file
need not even match what was on screen.

The accumulation now carries the microphone the way it already carried the protective
high-pass: frozen in `StartAsync`, as the resolved CURVE and the NAME its author was
shown, so neither a later rig change nor an edit to the calibration file behind it can
reach a capture already taken. Everything that draws or saves that capture reads the
frozen one; the rig's choice describes the next run.

The read-out follows the same rule from the other side: it names what the plot is
corrected through — a loaded capture by the name stored inside it, a running or held
one by its frozen calibration — and only an empty plot falls through to the rig. It is
a read-out, not a second selection, and it is refreshed wherever the plot changes
hands. The peak hold, likewise, is no longer dropped for changes that cannot touch the
curve it holds: only an SPL anchor can.

## The manual, read as a tuner rather than as its author

- **Sample rate**: forty-five lines of proof became twelve of instruction; the LR4
  numbers were already in `REFERENCE.md` and the two measured findings joined them.
- **The single-microphone comparison** claimed more than it should. Both figures are
  now the same measurement, smoothed and not: the difference falls to a median of
  1.3 dB under the width a single-point tune is read at, and the text says what an
  average actually buys instead of implying one microphone cannot tune.
- **The MMM pass** gained the rule that makes it valid — nobody in the acoustic path,
  which means the back seat — and the setting that decides its low end: Sequence
  Length at the maximum, because the excitation is one frame-length period of pink
  noise.
- **Cut**: five paragraphs of Excess Group Delay in a guide that never returns to it,
  the array's backend mechanics, the DSP processor's persistence model.
- **Corrected**: all-pass is a PEQ band; Auto delay's report leads with permission to
  skip it; the array is built AROUND the measurement microphone, which is not a row in
  its table and cannot be made one.

**Figures**: `measurement-options` and the manual's `record-settings` re-taken on the
current panel (it gained the *Measure through* row), `mmm-vs-single-point` re-taken
beside its new smoothed twin, and the array's own two figures.

## Verification

The full suite green — 159 Audio, 1686 App, 1338 Dsp — with new tests for the schema-12
migration, the panel carrying the rig's selection, a deleted calibration keeping its
place, the Own selection surviving a save and load, the frozen capture keeping its
calibration and its NAME across a rig change, the read-out staying read-only, and the
discard telling the shell. Each was checked against the unfixed code first. Two review
threads on this PR are resolved; one of them had become wrong once the freeze landed.


DIMOSUS:
Both fixed, and both were real.

**[P2] The analyzer was not told.** Right: `RefreshCalibrationConsumers` set `liveSpectrumOptions.CalibrationId` and stopped, so `RefreshCalibration()` — which invalidates the correction and drops the peak hold — was only reached from the old path, where the live panel owned the selection. The call now sits in `RefreshCalibrationConsumers` itself, which is where every path that can move the rig's choice passes (the live apply from Record Settings, and Apply), guarded on the id actually changing so an unrelated calibration-list edit does not drop a peak hold that is still valid. No test: that half lives in Form1.

**[P2] The read-out re-enabled itself.** Also right, and worse than described in one respect — it was not only after a refresh: `Configure` sets `Enabled = options.Count > 1` every time, so the very first list rebuild under an open panel undid the constructor's `Enabled = false`. Both callers go through one `ShowRigCalibration` now. Tested (`TheCalibrationReadOutStaysReadOnlyWhenTheListIsRebuilt`), checked against the unfixed code first — it fails there on the first assert.

Suite: 159 / 1693 / 1338.

DIMOSUS:
Blocker confirmed and fixed, and it was exactly as described — I checked it in the code before touching anything, since this came from a bot.

**The evidence.** `PlotModelFactory.BuildLiveCaptureDocument` takes the protective high-pass from `noiseMeasurement.CaptureProtectiveHighPass` (frozen on the accumulation, with a comment explaining why) and, two lines later, the calibration from `GetCalibration(liveSpectrumOptions)` — the live, mutable one. Same for every live render path. So the described sequence really does recompute run A's bins through B and stamp the file with B.

**The fix follows the existing model.** `NoiseMeasurement` now carries `CaptureMicrophoneCalibration` beside `CaptureProtectiveHighPass`, frozen in `StartAsync` at the same moment; every live render and the saved document read it. As you suggested, the resolved CURVE is frozen rather than the id, so editing or replacing the calibration file behind that id cannot reach a capture already taken. The controller gets the resolver injected rather than reaching through the plot factory.

**On your point 2 — you were right that `RefreshCalibration()` was the wrong fix.** With the freeze in place a held accumulation must not be re-rendered at all, so that call is gone from the rig-change path: the rig describes the next run, the capture keeps what it was taken through. Codex's symptom ("a stopped curve remains drawn with the previous correction") is the correct behaviour under this model, and `REFERENCE.md` now says so where MMM's frozen filter was already documented.

**Point 3** is fixed as described — one `ShowRigCalibration` that disables the box after every `Configure`, tested.

**Regression test**, your scenario verbatim: run with A, change the rig to B, build the document → the named calibration and the corrected curve are still A's. On the unfixed code it fails with `Expected: "cal-a" / Actual: "cal-b"`.

**CI**: agreed, the earlier run was cancelled rather than green. This push should give a full one; local suite is 159 / 1694 / 1338.

DIMOSUS:
Confirmed and fixed — with one deliberate difference from the suggestion.

The finding is right: `ApplyDisplayOptions` compares `CurrentPeakHoldKey()` against the rendered one and suspends the hold on any difference, and that key still carried `liveSpectrumOptions.CalibrationId` — the rig's choice — while the envelope is built from display points produced through the accumulation's frozen calibration. Choosing the next run's microphone mid-hold therefore dropped a valid envelope.

**Removed rather than repointed at `CaptureMicrophoneCalibrationId`.** That id cannot change while an envelope exists: it is written only in `StartAsync`, immediately after `SuspendPeakHold`, so by the time it moves the envelope is already gone. The protective high-pass is frozen at the same moment, is equally part of the display transform, and has never been in this key for exactly that reason — so keying on the frozen id would add a component that can never differ, which is the kind of state this branch has already deleted twice (the Phase/Group Delay calibration ids, and the live selector's copy).

No test: the key is private to the controller and reaching it needs a live Form, a PlotView and a running analyzer — which is why the peak-hold tests here poke fields on an uninitialized object rather than exercising the real path.

Local suite after the change: 159 / 1694 / 1338. CI running on the new head.

DIMOSUS:
Owner's requirement implemented, and it turned out to subsume the first finding.

**`Own (as measured)` is now an entry in the analysis calibration list**, and every new measurement selects it — finished sweep, opened file, history entry stepped back to. It names a rule rather than a curve, so unlike the `file-calibration` pseudo-entry it replaces it can be selected, persisted and restored, and it means the same thing against whatever measurement is open next session. Marked unavailable while the open measurement carries none, so "corrected by nothing" and "corrected by the microphone that measured it" do not look alike.

**That covers the P2 about a foreign file followed by a new sweep**, and covers it at the root rather than at the symptom: the suggested patch would have re-adopted the frozen curve only when the view still stood on `FileId`, leaving the general case — the view left on any of the user's own entries while the rig measured through another — untouched. That general case is the owner's own report: set *Measure through* to 90°, take a sweep, and the plot was drawn through 0° with the selector still naming it. Both are one bug, and `Own` is where it stops. `FileCalibrationSelection` and its displaced-selection bookkeeping are deleted with it.

**The peak-hold P2/P3 is fixed as you described**: `PersistCalibration` now refreshes the analyzer only when the SPL anchor moved. A microphone curve cannot reach a capture already taken, so dropping a valid envelope for it contradicted the freeze.

**Manual**: all three wordings taken — the run freezes into its result and *Save* writes the file; Group Delay is evidence rather than proof and only where level and coherence are worth reading; and the acoustic-path rule now says people and anything not part of the car, which is what it meant.

**Tab order**: fixed. Worth noting it predates this branch — the backend group has been ahead of the calibration rows since the array row was added in #130.

Local suite: 159 / 1686 / 1338. I will close the two stale inline threads once CI is green on this head.

DIMOSUS:
All three taken; the first two confirmed in code first.

**Imports (P2).** Real: `ImportRew` and the recorded-WAV path reach `ApplyLoadedImpulseResponseState` without ever passing through `AdoptRestoredResult`, so the view kept its previous selection and corrected an imported response with a microphone that had nothing to do with it. Fixed in ONE place rather than two: the reset moves into `ApplyLoadedImpulseResponseState`, which every from-disk path passes through — Resonalyze file, REW export, recorded WAV — and the history path keeps its own call, being the one that does not go there. No per-import regression tests: both paths live in Form1, which this repository does not test; the single shared call site is the argument that replaces them.

**Live read-out (P2).** Also real, and wider than described. Beyond `Init` being handed the rig's option, a loaded capture was not considered at all, so someone else's capture was labelled with this machine's setting. The controller now answers `DisplayedCalibration`: a loaded capture by the name stored in it, a running or held one by the id frozen when its run began, an empty plot not at all. The shell renders that and refreshes it at the four moments the answer changes — panel opening, capture load, run start/discard, calibration edit. The box also stopped being a combo bound to entry ids, because a loaded capture names a calibration from another machine's list which this one may not hold; it is a one-line read-out now.

**Sequence Length (P3).** Rewritten on the owner's instruction to say one thing: set it to the maximum, because the excitation is one frame-length period of pink noise, so the longest frame carries the most bass and the finest grid — 0.7 Hz bins at 48 kHz against 23 Hz at the default. The three metrics you separated are no longer mixed because only one is quoted.

Local suite: 159 / 1686 / 1338. This is one push rather than four — the earlier cancellations were mine, from pushing while a run was in flight.

DIMOSUS:
Both tails confirmed and fixed.

**The saved capture's calibration name.** Right, and it predates this branch: the document has always been `From(curve, <id>, fileName: null)` — before it was `liveSpectrumOptions.CalibrationId`, now it was the frozen id — so a capture taken through "90° capsule 2" has always carried `cal-<guid>` as its name. What is new is that the read-out surfaces it: while the capture is held the id resolves back through this machine's list, and after Save → Load it shows the guid verbatim. Fixed by freezing the NAME beside the curve and the id — `CapturedMicrophoneCalibration` carries all three, since resolving them at different moments is exactly how they come apart. The regression test freezes a generated id with a readable name and asserts the document keeps the name.

**Read-out after a discard.** Also right. Rather than a call beside each `DiscardCapturedData()`, the controller now notifies the shell itself, which is the robust half of your suggestion: the plot changes hands there, and that also refreshes Save availability — the same debt, since a discarded capture cannot be written. The controller test that half-builds the object with reflection now supplies that callback and asserts it fires.

Local suite: 159 / 1686 / 1338.

PR #136, opened 2026-08-29
Send a measurement to REW over its API

ayukhno:
Closes the second half of #90. `REW -> Resonalyze` already exists; this is the way back,
built to the decisions you set out in the thread.

**Where it is.** Right-click **Save** opens one item, *Send to REW...*. There is no menu
bar in the app, so this is the closest thing to `File -> Export`; say the word and it moves.
The item's caption reports what the availability probe found, and it stays clickable when
REW did not answer — the dialog behind it holds the address, which is the setting that
would fix that. Nothing is shown on success.

**Shape.** `source/Integration/Rew/`, a folder and not a project, three layers:
`RewImpulseResponsePayload` is pure (samples in, body out, base64 big-endian float32),
`RewApiClient` is the only `HttpClient` and takes it by constructor, `RewMeasurementExport`
sends and checks. No new package, no interface for one implementation, `dsp/` and `audio/`
untouched. One persisted setting, the base URL.

**Framing and level** as you decided: the buffer is rolled by whole samples (0.1 s, capped
at a quarter of the buffer) and the roll stated as a negative `startTime`, so t = 0 stays on
the loopback reference and the wrapped tail lands where it belongs; `splOffset` is sent only
when the measurement carries an SPL anchor, and it is the number the plot already uses
(`MeasurementPlotContext.SplOffsetDb` — the anchor plus this measurement's loopback level).

**Verification** does not gate on the REW version, as you asked. It reads the new
measurement's summary back and compares the arrival with the one the payload was framed to
produce, tolerating half a sample. A disagreement is reported in samples, with the version
REW announced in the text. The new measurement is found by a UUID that was not there before
rather than by title, because REW allows duplicates.

**Two measured things that change your plan slightly** (REW 5.40 Beta 132, API 0.9.6):

1. The summary's `timeOfIRStartSeconds` is **not** the start time that was sent — it is REW's
   own onset detection. A unit impulse sent at `-20.833 ms` reads back `0`; a band-limited
   arrival reads back six samples before its peak. The field that carries the start time back
   is `timeOfIRPeakSeconds`, because REW locates the peak by the same rule this application
   does, so the index cancels. That is what the check reads.
2. Omitting `splOffset` does not leave the import bare: REW fills in a default of its own
   (75 dB here), which looks like a calibration and is not — the same objection you made to
   the WAV route's 120 dB. Your instruction stands and nothing of ours is sent in its place;
   the dialog and `REFERENCE.md` say what the number is. Tell me if you would rather send an
   explicit `0`.

**Tests.** The payload layer and the whole conversation run against a fake handler, so CI
needs no REW. One live round trip is behind `[Trait("Category","Hardware")]` and a `RewFact`
gate on `RESONALYZE_REW_API_URL`; it creates one measurement and deletes it again. It is a
new attribute rather than `HardwareFact` because that one asks about WASAPI endpoints, which
say nothing about REW — happy to fold them together if you prefer one gate.

**Docs.** `REFERENCE.md` gains *Sending a measurement to REW* under Saving and Loading
Impulse Responses, including that an imported IR carries no timing reference and arrives
uncalibrated.

The suite is green, App/Audio/DSP, on a Windows runner.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1036,18 +1036,20 @@ covers a drag arriving while a sweep is running, or while a dialog is up: a file
 opened underneath one would replace the very measurement it is asking about.
 ### Sending a measurement to REW
 
-**Right-click Save** to open a short menu with **Send to REW...**, which imports the
-current measurement's loopback transfer function into a running
-[REW](https://www.roomeqwizard.com/) over its HTTP API. The menu asks REW whether it
-is there before it opens, so a REW that is not running says so in the item's caption
-instead of failing after the click.
+**Export** imports the current measurement's loopback transfer function into a
+running [REW](https://www.roomeqwizard.com/) over its HTTP API. The button is
+available exactly when **Save** is, minus the cases only this export cares about: a
+measurement with no transfer response, a sweep still running, and a moving-mic
+capture, whose Save belongs to that mode rather than to the impulse response this
+sends.
 
 The dialog names the measurement as it will appear in REW and holds the one setting,
 the address REW's API server listens on (`http://localhost:4735/` unless it has been
-moved; the server is switched on in REW's *Preferences -> API*). The address stays
-editable when REW did not answer, because it is the setting that would fix that.
-Nothing is reported when the send succeeds — the measurement is in REW, which is
-where you are looking.
+moved; the server is switched on in REW's *Preferences -> API*). REW is asked whether
+it is there as the dialog opens, so a REW that is not running is a line in the dialog
+rather than a failure after a click, and the address stays editable in that state
+because it is the setting that would fix it. Nothing is reported when the send
+succeeds — the measurement is in REW, which is where you are looking.
 
 **What arrives.** The transfer impulse response as measured, with t = 0 on the
 loopback reference. Its samples are sent unchanged and unresampled; the buffer is
@@ -1061,9 +1063,18 @@ reference of its own**: the shape and the internal delays are real, but REW cann
 compare its arrival with another REW measurement's. Level is sent only when the
 measurement carries an [SPL anchor](#sound-pressure-level-db-spl) — its own offset,
 combining the anchor with this measurement's loopback level. Without one, no offset is
-sent at all rather than a placeholder. REW then fills in a default of its own, which
-looks like a calibration and is not: nothing measured it, and the levels shown there
-are relative.
+sent at all rather than a placeholder. REW then fills in a default of its own (75 dB
+on 5.40 Beta 132), which looks like a calibration and is not: nothing measured it, and
+the levels shown there are relative.
+
+**And it carries no calibration**, which is worth knowing when one is selected here.
+An impulse response is stored raw — a [microphone calibration](#calibration)
+is applied when a curve is drawn, never baked into the samples — so what arrives in REW
+is the uncalibrated response, not the curve on the screen it was sent from. Resonalyze
+also tells REW **not** to apply a calibration of its own. REW would otherwise correct
+the arriving data by whichever calibration file is loaded in REW at the time, which
+belongs to REW's input rather than to the microphone that made this measurement. To see
+the calibrated curve in REW, load the same calibration file there.
 
 **What is checked.** After the import, Resonalyze reads the new measurement's summary
 back from REW and compares the arrival REW reports with the one the payload was framed
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -31,6 +31,7 @@ read-out refuses rather than guesses, what a number was measured against.
 - [Saving and Loading Impulse Responses](#saving-and-loading-impulse-responses)
   - [Importing a sweep recorded elsewhere](#importing-a-sweep-recorded-elsewhere)
   - [Dropping a file on the window](#dropping-a-file-on-the-window)
+  - [Sending a measurement to REW](#sending-a-measurement-to-rew)
 - [Plot Overlays](#plot-overlays)
   - [Target curves](#target-curves)
   - [Import and export](#import-and-export)
@@ -1033,6 +1034,43 @@ time — the shell holds one measurement, so a drag carrying several is refused
 while it hovers, the cursor showing that no drop will happen. The same refusal
 covers a drag arriving while a sweep is running, or while a dialog is up: a file
 opened underneath one would replace the very measurement it is asking about.
+### Sending a measurement to REW
+
+**Right-click Save** to open a short menu with **Send to REW...**, which imports the
+current measurement's loopback transfer function into a running
+[REW](https://www.roomeqwizard.com/) over its HTTP API. The menu asks REW whether it
+is there before it opens, so a REW that is not running says so in the item's caption
+instead of failing after the click.
+
+The dialog names the measurement as it will appear in REW and holds the one setting,
+the address REW's API server listens on (`http://localhost:4735/` unless it has been
+moved; the server is switched on in REW's *Preferences -> API*). The address stays
+editable when REW did not answer, because it is the setting that would fix that.
+Nothing is reported when the send succeeds — the measurement is in REW, which is
+where you are looking.
+
+**What arrives.** The transfer impulse response as measured, with t = 0 on the
+loopback reference. Its samples are sent unchanged and unresampled; the buffer is
+rolled by a tenth of a second so the deconvolution's acausal part, which is wrapped
+into the tail, appears before t = 0 where it belongs, and the roll is stated as a
+negative start time. A circular roll moves no energy, so the arrival still reads the
+delay this measurement found.
+
+**What it does not carry.** REW files an imported impulse response with **no timing
+reference of its own**: the shape and the internal delays are real, but REW cannot
+compare its arrival with another REW measurement's. Level is sent only when the
+measurement carries an [SPL anchor](#sound-pressure-level-db-spl) — its own offset,
+combining the anchor with this measurement's loopback level. Without one, no offset is
+sent at all rather than a placeholder. REW then fills in a default of its own, which
+looks like a calibration and is not: nothing measured it, and the levels shown there
+are relative.
+
+**What is checked.** After the import, Resonalyze reads the new measurement's summary
+back from REW and compares the arrival REW reports with the one the payload was framed
+to produce. A disagreement is reported in samples, together with the version REW
+announced. The REW version itself is deliberately not gated on — it is a moving beta,
+and pinning it would make matching builds your problem while proving less than this
+comparison does.
 
 ## Plot Overlays
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1053,10 +1053,14 @@ succeeds — the measurement is in REW, which is where you are looking.
 
 **What arrives.** The transfer impulse response as measured, with t = 0 on the
 loopback reference. Its samples are sent unchanged and unresampled; the buffer is
-rolled by a tenth of a second so the deconvolution's acausal part, which is wrapped
-into the tail, appears before t = 0 where it belongs, and the roll is stated as a
-negative start time. A circular roll moves no energy, so the arrival still reads the
-delay this measurement found.
+rolled by 150 ms so the deconvolution's acausal part, which is wrapped into the tail,
+appears before t = 0 where it belongs, and the roll is stated as a negative start
+time. A circular roll moves no energy, so the arrival still reads the delay this
+measurement found. The length is deliberate: REW builds its left gate out of the
+pre-roll and quietly clamps that gate to whatever pre-roll it is given. An import
+starts with a 100 ms left window in REW while REW's own swept measurements carry
+125 ms, so 150 ms is what lets an exported measurement be widened to the same gate
+as the REW measurement beside it, and compared with it like for like.
 
 **What it does not carry.** REW files an imported impulse response with **no timing
 reference of its own**: the shape and the internal delays are real, but REW cannot
--- a/source/Integration/Rew/RewApiClient.cs
+++ b/source/Integration/Rew/RewApiClient.cs
@@ -1,0 +1,192 @@
+using System.Net.Http.Json;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Resonalyze.Integration.Rew;
+
+/// <summary>
+/// The only place this application speaks HTTP to REW. Everything above it works
+/// in payloads and summaries, so the transport can be driven by a fake handler in
+/// tests and REW is never needed to prove the code that builds a request.
+/// </summary>
+/// <remarks>
+/// The <see cref="HttpClient"/> is supplied rather than created here for that
+/// reason; the shell hands it the one it owns.
+/// </remarks>
+internal sealed class RewApiClient
+{
+    /// <summary>Where REW's API listens when nobody has moved it.</summary>
+    public const string DefaultBaseUrl = "http://localhost:4735/";
+
+    private static readonly JsonSerializerOptions SerializerOptions = new()
+    {
+        PropertyNameCaseInsensitive = true
+    };
+
+    private readonly HttpClient http;
+    private readonly Uri baseAddress;
+
+    public RewApiClient(HttpClient http, Uri baseAddress)
+    {
+        ArgumentNullException.ThrowIfNull(http);
+        ArgumentNullException.ThrowIfNull(baseAddress);
+        this.http = http;
+        this.baseAddress = baseAddress;
+    }
+
+    /// <summary>
+    /// Turns a typed address into one that can be combined with a relative path.
+    /// Returns false for anything that is not an absolute http(s) URL, so a
+    /// mistyped setting is a refusal at the dialog rather than an exception later.
+    /// </summary>
+    public static bool TryParseBaseAddress(string? url, out Uri? baseAddress)
+    {
+        baseAddress = null;
+        if (string.IsNullOrWhiteSpace(url))
+        {
+            return false;
+        }
+
+        string trimmed = url.Trim();
+        if (!trimmed.EndsWith('/'))
+        {
+            trimmed += "/";
+        }
+
+        if (!Uri.TryCreate(trimmed, UriKind.Absolute, out Uri? parsed) ||
+            (parsed.Scheme != Uri.UriSchemeHttp && parsed.Scheme != Uri.UriSchemeHttps))
+        {
+            return false;
+        }
+
+        baseAddress = parsed;
+        return true;
+    }
+
+    /// <summary>
+    /// The version string REW announces, or null when it does not answer. Never
+    /// throws for an absent REW: not running is the ordinary case, not a fault.
+    /// </summary>
+    public async Task<string?> TryGetVersionAsync(CancellationToken cancellationToken)
+    {
+        try
+        {
+            RewApiMessage? message = await GetAsync<RewApiMessage>(
+                "version",
+                cancellationToken).ConfigureAwait(false);
+            return string.IsNullOrWhiteSpace(message?.Message) ? null : message.Message;
+        }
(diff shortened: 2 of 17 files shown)


DIMOSUS:
Checked it here before reading closely: Release build clean, and the App suite runs
1708 passed / 0 failed / 3 skipped with `Category!=Hardware`. The framing arithmetic holds
(the reference sample lands on `preRoll` and the negative start time puts it back on t = 0),
and so does the assumption the check rests on — `TransferFunction` takes the largest absolute
sample, which is REW's rule too, so the index really does cancel and the comparison is the
start time alone. The shape is what we agreed on: a folder and not a project, three layers,
the client by constructor, no new package, `dsp/` and `audio/` untouched, one setting, no
version gate.

Four things, then.

## 1. The surface: an Export button, not a right-click

You offered to move it, so: please do, and not into the Save/Load row. That row is two 102-wide
slots with a 6 gap, and `Compare`'s 210 is built on exactly that grid; a third button shrinks
all three to about 67 px for a niche feature, and the panel sits on fixed coordinates with a
font that autoscales, so a 67-px button is where the text clips at 150%.

Its own full-width row under `Compare` fits without argument: `(3, 120)`, `210 x 23`, and
`panel1` grows from `218 x 119` to `218 x 148`. There is room — the panel's bottom is at 265
and the next control below it starts at 309. Three of the five buttons there are already
full-width, so it reads as the same panel rather than an addition to it.

**The row is meant for two.** Import takes the seat beside Export when it arrives, and the row
splits into the Save/Load geometry: Export at `x = 3`, Import at `x = 111`, 102 each. Its first
tenant is the REW text import that is currently hidden inside `Load`, moved out onto a button of
its own; the API import you sketched earlier in this issue joins it there afterwards, as a second
entry behind that button rather than a third button. Until then Export spans the full width
rather than sitting alone next to a gap.

No menu behind it. One item is a click that buys the user nothing; let the button open the
dialog directly. The dialog already carries the probe's answer in its status line, so the probe
simply moves to the click — which also removes the up-to-two-seconds where a right-click on
Save looks dead at an address that does not refuse quickly. A menu can appear later, when there
is a second export to put in it.

Three requirements that come with a visible button:

- Declare it in `Form1.Designer.cs`, not in code after `InitializeComponent` — the repository's
  WinForms/DPI rule.
- Enable it exactly when `Save` is enabled: no transfer response, a run in progress, or MMM
  (`LiveCaptureOwnsSaveLoad`) must leave it **disabled**. Silently doing nothing was fine for a
  hidden gesture; for a button it is a defect.
- The `ContextMenuStrip`, the `buttonSave_MouseUp` handler, the `rewExportMenu` field and the
  added tooltip line all go. One surface, not two. `REFERENCE.md` changes with it.

## 2. `applyCal`'s comment states the opposite of what this application does

`RewImpulseResponseData` says the samples "already passed through this session's microphone
calibration". They have not. The transfer impulse response is raw: `ImpulseResponseFile`
documents it in as many words ("The impulse response is raw — no calibration is ever baked into
one"), the calibration is applied when a curve is rendered (`PlotModelFactory` ->
`RawCurveRenderer.CaptureCalibrationCorrection`, under `UseCalibration`), and the only place a
calibration is ever convolved into samples is the Virtual DSP audition dialog.

The consequence matters more than the comment. What lands in REW is not the curve the user is
looking at whenever a calibration is selected, and nothing says so: the dialog does not, and
`REFERENCE.md` does not either — the PR description says the new section covers that an imported
IR "arrives uncalibrated", but the section covers the timing reference and the level, and is
silent on calibration.

`false` is probably still the right value, since REW's calibration is not ours. But the reason
given for it is not true, and you are the one with a live REW to establish what `applyCal` does
before the comment is rewritten around it.

## 3. Nothing guards the form closing during a send

`ShowRewExportMenuAsync` checks `IsDisposed || Disposing` after its await; `SendToRewAsync` does
not. A send lives up to fifteen seconds in the filing poll, thirty in the transport, and the
form stays interactive throughout, so closing the application inside that window leaves
`MessageBox.Show(this, ...)` reaching for a disposed form's handle. Narrow, but it turns a
message into a crash, and the guard already exists two methods above.

## 4. Your three questions

**`timeOfIRStartSeconds`** — accepted, and it corrects what I asked for: I said to compare
`startTime` and `timeOfIRPeakSeconds`, and the first of those has nothing to compare against.
The peak field is the right one for the reason you give.

**`splOffset`** — no, do not send an explicit `0`. Zero is also a claim, and it says "these
numbers are dB SPL". Omission is the honest answer; the dialog and `REFERENCE.md` already say
what REW will invent in its place.

**`RewFact`** — keep it separate. A WASAPI endpoint says nothing about REW, and folding the two
gates together would run or skip these tests for the wrong reason.

DIMOSUS:
Three more things I found while checking the current head. These are separate from the four points above.

## 5. The probe timeout can escape as an unhandled cancellation

`ShowRewExportMenuAsync` creates a `CancellationTokenSource(RewProbeTimeout)` and passes that token into `ProbeAsync`. But `RewApiClient.TryGetVersionAsync` treats `OperationCanceledException` as "unreachable" only when `!cancellationToken.IsCancellationRequested`:

```csharp
OperationCanceledException => !cancellationToken.IsCancellationRequested
```

When the *two-second probe timeout itself* fires, that token is of course cancelled, so the exception is not converted to `null`; it propagates out. Today that reaches `buttonSave_MouseUp`, which is `async void`, so an address that silently drops packets rather than immediately refusing the connection can turn the intended `(not answering)` state into an unhandled UI exception.

The existing `Unreachable` fake only throws `HttpRequestException`, so it does not cover this case. Please make the short probe timeout an expected "not answering" result and pin it with a cancellation/timeout test. This still applies after moving the probe onto the new Export button.

## 6. The live-test cleanup can delete a user's measurement

`DeleteMeasurementsAddedSinceAsync` currently deletes every UUID that was absent from the snapshot taken before the test. That is broader than "delete what this test added": if the user's REW is genuinely mid-session, as the comment explicitly allows, and they make a measurement while the test is running, cleanup will delete that measurement too.

The test already gives its own import a unique identifier (`Resonalyze round trip {Guid...}`). Please identify and delete only the measurement created by this test, not every measurement added concurrently.

## 7. The same new-UUID race exists in production verification

`WaitForNewMeasurementAsync` returns the first UUID that was not present before the POST. If the user creates a REW measurement between the `before` snapshot and the point where this import is filed, Resonalyze can pick that unrelated measurement and verify its peak time instead.

UUID is still the right way to distinguish this send from an older measurement with the same title, but title can be used as an additional discriminator: require both `UUID not in before` **and** `Title == request.Identifier`. That preserves duplicate-title safety while excluding unrelated measurements created concurrently, and it gives the live test a safe target to clean up as well.

ayukhno:
**All seven are in, and one of them needed your instinct about it being unverified — it was.**

Rebased onto current `main` first, so this sits over #153. Fork CI on Windows: App 1988 passed / 0 failed / 3 skipped, Audio 159, DSP 1410.

## 1. The button

Export now has its own full-width row under `Compare` at exactly the geometry you gave — `(3, 120)`, `210 x 23`, `panel1` grown to `218 x 148` — declared in `Form1.Designer.cs` rather than added after `InitializeComponent`. The `ContextMenuStrip`, the `buttonSave_MouseUp` handler, the `rewExportMenu` field and the tooltip line are gone; `REFERENCE.md` changed with them. No menu behind it: the click opens the dialog and the probe moved onto it, which also removes the up-to-two-seconds where a right-click on `Save` looked dead.

Availability is set in `MainCommandController.SetSaveAvailable` rather than beside each caller, because there are eight callers and a button that has to track another one cannot be left to each of them to remember. Export is offered exactly when `Save` is, narrowed by what only it cares about. A run in progress needs no new wiring — `EnterMeasurementRunningState` already calls `SetSaveAvailable(false)`.

## 2. `applyCal` — measured, and it does something

You were right that the comment stated the opposite of what happens, and right that this needed a live REW rather than a rewrite.

Method: post a synthetic Dirac rather than a real measurement. A flat spectrum means whatever REW draws **is** the correction, read directly instead of inferred. The same impulse imported twice, `applyCal` true and false, against 5.40 Beta 132 / API 0.9.6:

| | 20-500 Hz | 1 kHz | 4 kHz | 10 kHz | 16 kHz | 20 kHz |
|---|---|---|---|---|---|---|
| ON - OFF | +0.000 | -0.311 | -0.236 | +1.081 | +2.800 | +3.900 |
| -cal(f) | +0.009 | -0.302 | -0.227 | +1.089 | +2.808 | +3.837 |

`ON - OFF = -cal(f)` with a fitted constant of 0.009 dB and a worst error of **0.06 dB** across the band — a subtraction in dB, which is the linear magnitude divided by the calibration. Two traps for anyone repeating it: the cal must be loaded **before** the import, and it must sit on the entry REW shows as active.

So `false` stays, for your reason rather than the written one: the curve REW would apply belongs to REW's own input, not to the microphone that made the measurement. The comment now says that, and `REFERENCE.md` has a **What it does not carry** paragraph on calibration, which it was missing. Incidental from the same run: with `splOffset` omitted, REW invents **75 dB**, so the docs name the number now instead of gesturing at it.

## 3. The form closing during a send

`ReportRewProblem` is a no-op on a disposed form, and the wait cursor is only put back on a form still there.

## 4. Your three answers

All three taken as given.

## 5. The probe timeout

`ProbeAsync` takes the timeout and cancels a linked token of its own, so its own expiry is "not answering" while a caller's cancellation still propagates. Pinned by two tests against a fake REW that accepts the connection and then says nothing — the case the refusing fake never covered.

## 6 and 7. The new-UUID race, in both places

Both now require `Title == identifier` alongside the new UUID, which keeps the duplicate-title safety the UUID was there for. The live test deletes only its own import.

---

**Still open from #88:** the LR36 electrical measurement is next on the rig.

DIMOSUS:
**One correction, and the mistake is mine.** When I gave you the geometry I said the panel's
bottom was at 265 and the next control below it started at 309. That skipped `buttonHistory` — a
form-level control at `(1268, 280)`, `210 x 23`, sitting in the gap I called empty. The room was
15 px, not 44, and a 23-px row does not fit in it.

So on the current head the rectangles overlap:

- `panel1` now spans 146-294 (it spanned 146-265)
- `buttonHistory` spans 280-303
- Export, inside the panel, spans 267-290 — the panel's border puts its client origin at 147

Fourteen pixels of overlap with the panel, ten with the button itself. And `buttonHistory` is
added to `Controls` before `panel1`, so it holds the lower index and draws in front: the bottom
ten pixels of Export are covered by History. Both are anchored Top | Right and scale together, so
it holds at every window size and every DPI.

Moving the two buttons below the panel down by the height of the row it gained keeps every
existing gap intact:

- `buttonHistory` 280 -> 309 (the 15 px under the panel, as before)
- `buttonCurrentModeSettings` 309 -> 338 (the 6 px under History, as before)

`buttonCurrentModeSettings` then ends at 361, and the next control down, `buttonOverlayShowAll`,
is at 453 — 92 px of slack left over.

Nothing else from the round needs revisiting. I rebuilt the branch here: App runs 1988 passed /
0 failed / 3 skipped, and all seven answers check out — including the `applyCal` measurement,
which was the right way to settle a question a rewrite could not have.

DIMOSUS (commented):
Two things left from the current head.

### 1. `/measurements` JSON failures are not translated into a REW-facing error

`GetMeasurementsAsync()` lets `ReadFromJsonAsync` throw `JsonException`, while `SendToRewAsync()` only handles `RewApiException`, `HttpRequestException`, and `TaskCanceledException`. So if a future REW beta still answers but changes or malforms the `/measurements` JSON shape, the export falls through to the application's global `unexpected error` handler instead of reporting a normal "REW returned an unexpected response" problem.

That is especially relevant here because the design deliberately avoids a version gate. I would wrap JSON-shape failures from `GetMeasurementsAsync()` in `RewApiException` (and probably `NotSupportedException` too) and pin it with a malformed-`/measurements` fake test.

### 2. The promised round-trip still does not demonstrate magnitude / phase / level

The live test proves that the import route works and that `timeOfIRPeakSeconds` lands where expected, which is good. But the validation we agreed on in #90 was numerical `REW -> Resonalyze -> REW` preservation of magnitude, phase, relative/absolute level, and timing. The current hardware test sends `splOffsetDb: null` and only checks the timing result, so level never goes through a live REW and magnitude/phase are not checked at all.

I do not think the full eight-channel harness needs to move into CI. But before merge, please post the promised measured results from the published set: max magnitude error, max phase error, level error, and timing error. The phase check is particularly useful here because the payload circularly rolls the IR and relies on the negative `startTime` to preserve the original time base.

ayukhno:
**Both are in, and the round trip found two things before it produced a number —
one of them a defect in what the last round added.**

## 1. The button row

Your correction is right, and it reproduces from the file rather than the screen:
with Export in it `panel1` spans 146-294, `buttonHistory` spans 280-303, and
History holds the lower index, so it draws over the bottom ten pixels of the new
row. Moved exactly as you set out — `buttonHistory` 280 -> 309,
`buttonCurrentModeSettings` 309 -> 338. The latter now ends at 361 and
`buttonOverlayShowAll` is still at 453.

## 2. The `/measurements` shape

Wrapped, with four tests on the shape — malformed body, an HTML page, an unreadable
charset, and the same failure on the poll AFTER the import has gone through — and one
on the probe. One correction to the suggestion, measured rather than assumed
(.NET 10.0.301):
**`NotSupportedException` is not what this route raises.** The generic
`ReadFromJsonAsync` does not reject a foreign content type at all — it parses the
bytes anyway, so REW's own HTML 404 page arrives as a `JsonException` about `'<'`.
The type that *does* come from the header is `InvalidOperationException`, raised
when the Content-Type's charset cannot be resolved, and no `JsonException` catch
would have covered it. Both are caught; `ObjectDisposedException` is excluded
because it derives from the second and means this application misused its own
`HttpClient` rather than anything about REW. The version probe now shares the rule:
something on 4735 that answers and is not REW is the same news as nothing
listening, and must not throw at the button.

## 3. The round trip, on the published eight-channel set

Each channel of `2026-08-20_front-set-02` sent from its published v7 file through
this exporter's own code, read back from REW, and compared with what was sent.

**Two findings first, because both changed the answer.**

**Ours: the pre-roll was too short for REW's gate.** REW builds its left gate out
of the pre-roll and *clamps* that gate to the pre-roll present, silently. An import
starts with a 100 ms left window while your swept measurements here carry 125 ms —
so at the old `PreRollSeconds = 0.1` an exported measurement could not be given the
same gate as the REW measurement beside it (asking for 125 ms yielded 100.4), and
comparing the two compared REW's windowing as much as the responses. On a
nine-measurement crossover bench that alone read as 0.18 dB rms and 0.94 degrees
rms. Now 0.15 s.

**Mine to have introduced, last round: REW truncates the title.** A 54-character
name comes back at 48, a 64-character one at 45 — a display width, not a character
count. The `Title == identifier` check therefore never matched for a long name: the
export waited out the fifteen-second filing timeout and told the user REW had not
filed the measurement, while REW had filed it perfectly. Now matched as a prefix,
in the verification and in the live test's cleanup — which had been walking past
its own measurement and leaving it in the session. The pure suite could not catch
this; running it live did.

**The numbers**, per channel, over 20 Hz - 20 kHz on bins carrying signal:

| channel | startTime d | peak index d | samples, of FS | mag rms | mag peak | phase rms | phase peak |
|---|---|---|---|---|---|---|---|
| c    | 0.000e+00 | 0 | 5.2e-08 | 6.0e-07 dB | 1.2e-05 dB | 4.2e-06 deg | 6.1e-05 deg |
| m_L  | 0.000e+00 | 0 | 3.9e-08 | 4.4e-07 dB | 1.0e-05 dB | 3.4e-06 deg | 8.1e-05 deg |
| m_R  | 0.000e+00 | 0 | 3.8e-08 | 3.9e-07 dB | 8.0e-06 dB | 2.8e-06 deg | 7.0e-05 deg |
| sw   | 0.000e+00 | 0 | 6.0e-08 | 1.6e-05 dB | 5.4e-04 dB | 1.0e-04 deg | 3.0e-03 deg |
| tw_L | 0.000e+00 | 0 | 3.5e-08 | 7.1e-07 dB | 1.6e-05 dB | 4.2e-06 deg | 1.0e-04 deg |
| tw_R | 0.000e+00 | 0 | 4.0e-08 | 3.9e-07 dB | 1.0e-05 dB | 2.4e-06 deg | 6.7e-05 deg |
| w_L  | 0.000e+00 | 0 | 6.0e-08 | 1.9e-06 dB | 1.4e-05 dB | 1.2e-05 deg | 1.2e-04 deg |
| w_R  | 0.000e+00 | 0 | 5.6e-08 | 1.5e-06 dB | 1.6e-05 dB | 1.1e-05 deg | 1.3e-04 deg |

`startTime` and the peak index are exact on all eight. The samples come back within
6e-8 of full scale — half a float32 ULP — at every one of the 262144 indices, so
nothing is shifted or resampled. `splOffset` returns exactly as sent on all eight.

**Inter-channel level holds to 0.003 dB across seven of the eight.** The eighth is
the subwoofer, and the gap there is REW's gate, not the transport: the constant
relating sent samples to the SPL REW reports moves monotonically toward the other
seven as the window is opened — **-8.88, -6.29, -5.09, -3.79 dB** against their
**-3.010** — and cannot reach them because a gate never covers the whole 2.73 s
buffer. The sample comparison, which no gate touches, is the same 6e-8 for the
subwoofer as for every other channel.

**Two things this does not establish, stated so the table is not read for more than
it says.** REW's own `/impulse-response` hands back a peak-normalised copy, so the
samples had to be compared after removing that one scalar — the scalar is REW's, on
the way out, and the export never sees it. And the published files carry no
Resonalyze SPL anchor of their own, so the offset used per channel is the one REW
recorded; that makes this `REW -> here -> REW` for level, not a test of how the
anchor is derived on our side.

The eight-channel harness stays out of CI, as you said. The live round-trip test in
the suite is still one measurement, still gated on `RESONALYZE_REW_API_URL`, and now
passes in about a second instead of timing out.

**One check on what you asked for, before you read the table for more or less than it
is.** I took "the published set" to mean `2026-08-20_front-set-02` — the eight channels
we published and promised these numbers on in #90 — and ran the round trip from those
stored v7 files, so the input is something you already have and can re-run against. If
you meant a fresh capture instead — a new measurement taken here and pushed straight
through, so what gets exercised is the live path out of a running sweep rather than a
file — say so and I will do that. It is a rig session rather than a re-analysis, so it
would come a little later.

DIMOSUS:
Checked `5977dec4` and the latest round. The previous review points are closed well, and the 8-channel round-trip numbers look excellent. The 150 ms pre-roll change is also justified by the live REW behavior you found.

One blocker remains before merge: the new prefix match re-opens the concurrency race. `identifier.StartsWith(filedTitle)` means a concurrent new measurement named e.g. `Resonalyze` can match an export named `Resonalyze 2026-09-01 ...`. `WaitForNewMeasurementAsync()` may then verify the wrong UUID, and the live-test cleanup can again delete a user's measurement if its title happens to be a matching prefix. The current concurrent-measurement test does not cover that case. Please use a discriminator that remains unique after REW truncation (for the live test, putting the GUID at the beginning would already make cleanup safe), and make Export single-flight so a second click cannot start another probe/send concurrently.

Small robustness follow-up: `ImportImpulseResponseAsync()` still catches only `JsonException` while reading a non-success POST body. The same unreadable-charset `InvalidOperationException` now handled for `/measurements` can still escape here; reusing the new unreadable-response classification would make the transport consistent.

Also worth refreshing the PR description before merge: it still describes right-click Save, 0.1 s pre-roll, and UUID-only matching.

DIMOSUS:
Checked the latest head. The race fix, single-flight guard, live-test cleanup discriminator, and unreadable POST response handling all look good. CI is green as well.

I don't have any remaining merge-blocking concerns. From my side this is ready to merge.

PR #147, opened 2026-08-30
Give every channel block a zone, and room for twelve of them

DIMOSUS:
First step of the "complex installations" work, from the reference session
`Monster` — seven blocks: a front three-way, a rear pair, a mono centre and
two mono subwoofers in different bands.

## Why

A complex install is not one chain along the spectrum, which is the only
shape the tool could describe. Rear and centre drivers routinely play the
SAME band as the front midrange and tweeter, from another place, with no
crossover junction to anyone. Ordering blocks by band centre — all the tool
could do while every block was just "a channel" — then invents a handover:
on the reference session it declares a `Mid ↔ Rear` junction at 3500 Hz,
which is the front midrange's low-pass and has nothing to do with the rear
fill. Auto delay does not solve that badly; it solves a question nobody
asked.

This PR adds the fact the tool was missing, and nothing that consumes it
yet — the grouped views, the staged alignment and the per-group crossover
wizard follow in their own PRs. **No engine is touched here**, so the
session battery is unchanged by construction.

## What

**A zone per block** — Front, Rear, Center or Sub — as a combo box beside
the Mono checkbox rather than replacing it. The two are independent in the
field, which the reference session itself proves: it carries two mono subs
in different bands, and a sub pair can legitimately be stereo while a
two-way centre is two mono blocks. Merging them into one list would force a
re-typing whenever a mono sub becomes a stereo pair. Centre is the one zone
that implies mono, and the card enforces it (checked and locked; leaving
Center releases the lock without clearing the box, so re-typing a
mis-guessed centre as a sub does not silently split a shared driver).

Sub is a zone for the views and the tuning sheet. For alignment the
subwoofers stay the bottom of the front chain, which is where their
junctions actually are — on the reference session the two subs form real
handovers with each other and with the midbass.

**Schema v9**, with a migration that is purely additive. A v8 file records
no zone, so it is guessed from the two facts it does record: a stereo pair
becomes Front (a rear pair reads identically in a v8 file), and a mono
block becomes Sub unless it HIGH-PASSES, which no subwoofer does — that one
is a centre. Every delay, gain, polarity, crossover corner, PEQ band, gate
and view flag survives untouched, which is what makes a wrong guess cheap:
the project opens exactly as it was tuned and the user re-points one combo
box.

The real reference session migrates with six zones of seven right — the
centre and both subs land correctly, and only the rear pair needs the one
manual change the guess cannot avoid.

**Card layout**, re-ordered as the owner asked: the curve toggles move up
under the source button, the zone row takes the polarity row's place, and
the delay's millimetre readout moves into that field's tooltip to make
room. The fold line follows the crossover row on its own (it is read from
`comboBoxCrossoverKind.Top`), so folding still cuts above the filter chain —
with the bonus that a folded block now shows its Raw/Processed toggles,
which is exactly what a user reaches for while comparing blocks. The block's
height is unchanged at 206 px; a folded one grows from 74 to 99.

**Ceiling raised from 8 to 12 blocks**, with four palette colours added
(terracotta, slate blue, sand, teal). Seven blocks is the reference session
today, so splitting the rear or the centre into two ways used to hit the
wall.

## One measurement that contradicted the plan

The right-hand metric read-out was expected to just need moving up. It
does not fit either way, and it **already overflowed before this branch**.
Worst case (all blocks stereo, all three sections present — sum loss,
junction phase, stereo deltas), Consolas 9 pt at the label's width:

| Blocks | Needs | Label was |
|---|---|---|
| 6 | 520 px | 431 px |
| 8 | 632 px | 431 px |
| 10 | 744 px | 431 px |
| 12 | 856 px | 431 px |

The window's minimum is 1494×832, so even at the very top of the column
there are only 820 px — twelve blocks cannot fit by moving alone. Virtual
DSP hides the whole capture block, so the column below the warning box is
free: the label moves to y=150 with height 670, which covers 6 and 8 blocks
at the minimum window size and 10–12 on any larger one (it is anchored top
and bottom).

The complete answer is a scrollable read-out, deliberately deferred to the
grouped-views PR: those print ONE group rather than every junction in the
system, so they shorten this text themselves. Deciding on a scrollbar makes
sense once the new length is visible, not blind now.

## Tests

- Zone/mono interaction on the real control: Center forces and locks Mono,
  leaving Center releases without clearing, every other zone stays free.
- The legacy classification, pinned per branch.
- v8 → v9 round trip asserting every tuned setting comes back identical.
- A zone outside the enum is rejected rather than read as Front (the name
  cannot deserialize, but a NUMBER would land silently outside it, and a
  block in no zone would vanish from every grouped view without a word).

Two fold tests were **recalibrated, not adjusted**: both had their subject
changed by the layout. One asserted Bypass is hidden when folded — it is
now above the fold, which is the requested behaviour, so it is re-pointed
at the PEQ row (genuinely the bottom one). The other required the fold
button to come after Invert in the tab order; it is now that row's leftmost
control, so the test is renamed and checks the whole row's order instead.

Full suite green: 3390 passed, 0 failed.


DIMOSUS:
Both findings confirmed against the code and fixed in 4f712fd.

**P1 — the tooltip claimed calculations it does not drive.** Correct, and it is the worse of the two: nothing reads `Pair.Zone` outside the panel binding and the migration, so a user marking an overlapping driver as Rear would have believed the fictitious `Mid ↔ Rear` junction was gone while it was still being formed by band-centre ordering. The tooltip and the enum's doc comment now say the zone labels the block and travels with the project, and name the changes that will act on it. (The PR body already said no consumer exists yet; the tooltip a user actually reads did not, which is the part that mattered.)

**P1 — the manuals had gone false.** Correct, and it is an explicit repo rule I broke: AGENTS.md requires the documentation in the same commit and says to grep all three files for the terms the change makes obsolete. Done —

- `README.md`: eight pairs → twelve.
- `REFERENCE.md`: eight → twelve; the delay's "live **mm** read-out" now describes the tooltip it moved into; and a new **Zone** entry covering the four values, why zone and Mono are separate fields, Center's forced Mono, the honest current scope, and the legacy guess (so a user can explain the zones an old project opens with).
- `MANUAL.md`: one line in the Virtual DSP setup step, since setting the zone is now part of what the tuner does.

Re-grepped the three files afterwards; the remaining "eight" hits are the microphone array and the EQ band fit, unrelated.

**@DIMOSUS — one figure needs re-taking: `assets/images/manual/virtual-dsp.png`.** I first wrote here that it did not, then opened it and found the opposite: at 1720×1035 the channel blocks are fully legible, and the shot shows the old row order (Curves at the bottom of each block), the `= 0 mm` / `= 885,5 mm` labels that this PR removes, and no Zone selector. The right-hand read-out also starts lower than it now does. Per AGENTS.md that is a `tools/Resonalyze.Screenshots` re-run on your own measurements. Correcting the claim rather than editing it away, since it was wrong in a direction that would have cost you a stale figure.

Full suite still green: 3390 passed, 0 failed.

DIMOSUS:
All three points addressed in 4ad2451.

**P2 — figure re-taken.** I found the session it came from (`D:/hobby/AMP/v5_exp/`, the four measurements the shot shows) and re-ran `tools/Resonalyze.Screenshots -- manual/virtual-dsp`. Three things worth knowing:

- The annotation box for region **6** had to move with it. Those boxes are hard-coded figure pixels in `Shots.cs`, so the read-out moving up the column left its frame behind at the old position — visible as a border starting halfway down with the text above it. Now `Box(1490, 145, …)`.
- The session on disk (`virtual-dsp-session-manual-3.json`) has `activeSideRight` and the **coherence** junction view, but the prose beside the figure says region 5 shows *"the correlation and score curves"*. I shot it with those two view flags flipped back so the figure and the manual agree — the tune itself is untouched. My local `screenshots.json` carries a note about it, since a plain re-run from that session would silently disagree with the text again.
- The zones in the new figure (**Sub / Front / Front / Front**) are the v7→v9 migration running for real, which is a better demonstration than anything I could have written.

**P3 — the delay tooltip bug is real, and worse than described.** Confirmed: my earlier edit to `ApplyTooltips` silently failed to apply, so the static text survived — and that text names an "mm readout" which, after this branch, is the tooltip itself. A freshly built block therefore described a control that no longer exists until the next delay edit. Fixed, and pinned by two tests covering both installation orders; the one that was broken is the one every real block takes. Worth noting this is the second time in this PR that a claim of mine outran the code, both caught by review rather than by me.

**P3 — stale XML summary** on `VirtualCrossoverPanel` ("up to eight measured transfer IRs") corrected to twelve.

**Plus, at the owner's request:** the tooltip now gives the distance in millimetres *and* inches on one line — `885,5 mm (34,9 in)`. Both always, rather than a unit setting: which tape the reader owns is not something the app can know, and a setting would leave the other half of the world doing arithmetic at the moment they wanted a figure to compare. `REFERENCE.md` says both units too.

Full suite green: 3392 passed, 0 failed.

PR #148, opened 2026-08-30
Let the plot ask about one part of the installation at a time

DIMOSUS:
Second of the "complex installations" series, on top of #147, which gave every
block a zone but had nothing read it. This is what reads it.

## Why

The front three-way, the rear pair and the centre of the reference session all
cover 290 Hz upward from different places. Drawn together they are seven
overlapping traces whose sum describes no listening position — and whose
junction read-outs describe crossovers that are not in the car, because the
tool pairs channels that are adjacent along the spectrum and there is no filter
handing a band from a front midrange to a rear fill.

## What

A **Show** selector under the plot picks a subset that is a coherent question:

| Show | draws | sums |
| --- | --- | --- |
| **Front + Sub** | the front stage and the subwoofers | all of them |
| **Rear + Sub** | the rear fill and the subwoofers | all of them |
| **Front + Center** | the front stage and the centre | the front stage only |
| **Groups** | one summed line per zone, no driver curves | each zone separately |
| **Everything** | every driver | everything but the centre |

Front + Sub is the default and the enum's zero, so a project with no rear or
centre never meets any of the rest — it draws exactly what it always drew.

**The choice is made once, at the top of the frame.** The curves, the sum, the
loss trace, the read-out and the opposite side's comparison curve all read the
same filtered set. Four places deciding separately is how a plot ends up summing
a channel it does not draw.

## Two things withheld rather than computed wrongly

**A centre never enters a sum, in any view.** It plays a signal synthesised from
L and R — the correlated content pulled out of both — so how much of the
programme reaches it is a property of the track, not of the tune. Its measured
response is an honest acoustic path, but adding that path to the front stage's
would state a division of signal between them that no measurement can know and
that changes from one recording to the next. It is drawn to be compared.

**A view spanning more than one group quotes no summation loss and no junction
phase.** Both describe handovers. Front against rear is not one: they play the
same band from opposite ends of the cabin with no filter between them, so their
sum combs however well either is tuned, and the figure would report several
decibels of damage that nothing can repair on a system that is correct. The
junction-phase block is worse than useless there rather than merely long — it
builds its pairs by band adjacency, so on the reference installation it would
have printed a row for a front midrange handing over to a rear fill at 3500 Hz.

What those views report instead is the pair of numbers a tuner actually sets
between groups: a **vs Front** block giving each group's arrival difference
(positive = later, which is what a rear fill wants) and level difference,
measured on the groups' summed responses in the band they share. It sits
directly under the sum-loss column because it is what stands in for it. A test
pins the invariant: every view either quotes a loss or compares across groups,
so none of them is silent.

The **Groups** view draws one summed line per zone and no driver curves, all
gated on one shared anchor rather than per group — the whole point is to read
the groups' relative timing off the plot, and per-group anchors would each hide
their own group's delay by construction.

## Notes

- `BuildCurves` now takes the drawn set and the summed subset separately, sharing
  one window anchor. A sum anchored differently from the curves it is drawn over
  stops being their vector sum, and the loss can poke above its 0 dB ceiling.
- An empty view (a rear view of a front-only car) says so and names the fix,
  rather than falling back to drawing the whole system.
- The cross-group block reads the responses the frame already processed, so it
  adds no render — only the arrival FFTs, on the coordinator's auxiliary path
  with the same silent-drop-when-superseded behaviour as the stereo deltas.
- The new controls are registered in the panel's manual `plotFollowers` layout
  pass. Without that they keep their designer Y while everything else moves with
  the plot, which put the selector on top of the graph — caught in the re-shot
  figure.
- `REFERENCE.md` gains a Show section, and the Zone entry's "nothing reads this
  yet" paragraph is now half false, so it says what does and what still does not.
  `MANUAL.md` says a front-only car can set Front and Sub and forget it.
- The figure is re-taken; the region-6 box needed no move this time.

## Deferred question from #147, answered

That PR left open whether the read-out needs a scrollbar. Withholding the
junction-phase block across groups removes about fourteen lines from the worst
case, which is what the overflow was made of — the everyday views now fit the
670 px label comfortably. **Everything** on a full twelve-block project is the
one case still close to the edge on a minimum-size window; it is also the view
nobody tunes in. Recommend leaving it, rather than adding a scrollbar for it.

Full suite green: 3412 passed, 0 failed.


DIMOSUS:
All four confirmed against the code and fixed in 470b651. The first three are the same mistake in three places — drawing one set and measuring another — which is the mistake this PR exists to prevent, so thank you for catching them.

**P1 — Front + Center junction metrics included the centre.** Correct, and it makes the PR body's own claim false: I wrote that the loss "still describes the front chain alone", but `BuildEntries`/`BuildPhaseEntries` were handed `shown`, so the centre was paired with its neighbouring front driver as a junction. Both now read `summedChannels` — the set the loss curve was actually divided out of.

**P1 — the phase Sum included the centre.** Correct. `BuildAcousticRender` now takes the summed subset alongside the drawn set and passes it to `BuildPhaseCurves`, whose Sum gates and combines only those spectra; it also withholds the Sum below two summing channels rather than emitting a "sum" of one.

**P2 — Groups bypassed Hybrid and the target.** Correct on both, handled differently because they differ in kind:
- The **target** is now drawn in the Groups view. Judging a rear fill's level against the house curve is half of what that view is for, so its absence was a plain omission.
- **Hybrid** is muted there instead of implemented. A spatial average is a property of one DRIVER, and this view draws only group sums — there is nothing honest to hang one on. The toggle now greys with a tooltip saying exactly that, alongside its two existing muted cases, and the view switch re-judges availability.

**P2 — an empty view left the previous warning standing.** Correct; `HideWarning()` on that path.

Added a test that pins the first one as an invariant over the views rather than a fact about one call site: a view that quotes a summation loss and also draws something it does not sum must build its junction rows from the sum. Front + Center is the only view in that corner today, and the test names it — if another joins, that is the reminder.

`REFERENCE.md` notes the target and the muted Hybrid in the Groups view.

Full suite green: 3413 passed, 0 failed.

DIMOSUS:
All three confirmed and fixed in c83ea3a. The first one is the sharpest catch of the round — thank you.

**1. Rear + Sub reinvented the very defect the PR removes.** Confirmed. Subwoofers stopping at 110 Hz beside a rear fill starting at 290, and `GetAdjacentPairs` pairs neighbours in the band ordering without asking whether anything crosses between them. So the view that killed an invented *front midrange → rear fill* printed an invented *sub → rear fill* instead.

Your architectural note was the useful part: `LossChainZone` returns a zone that the caller only tests for null, and the returned zone constrains nothing. I did **not** fix it by using that zone, though — filtering junction members by zone would be wrong for the front chain, where the subwoofers genuinely do hand over to the midbass and belong in it. The rule is about the handover, not the zone: a junction now requires both channels to actually PLAY inside the octave-each-way window the junction would be measured over. That reuses the measurement's own window rather than inventing a gap tolerance, and it keeps every real handover — including drivers deliberately crossed a third of an octave apart, which is pinned by a test beside the gap case.

Worth noting it also fixes the same latent bug outside the grouped views: any project with a genuine hole in its chain was getting a junction row for it. The full suite is unchanged by the rule, because a real chain has real handovers.

**2. Groups only behaved as Groups in Magnitude.** Confirmed — the phase and impulse branches return before the group-sum branch. Took your first option: Groups is a magnitude view, so picking it moves the view radio to Magnitude and mutes Phase and Impulse, visibly, rather than drawing what the selector denies.

**3. Hybrid was only muted, not disabled.** Confirmed, and the earlier fix was skin deep — `HybridRequested` read only the tick and the coverage, so a hybrid ticked before the switch went on being computed and the read-out reported a spatial-average offset for curves that were not on the plot. The view is part of the condition now. Sum and Sum loss are muted where their answers are ignored.

One thing your report did not name that fell out of testing it: a project that OPENS on a grouped view refreshed none of this, because the selector's event is suppressed while the state is applied — it reopened with every muted control bright and inert. Fixed on the load path too.

Verified by rendering the Groups view on a real session rather than trusting the compiler: two zone sums drawn, target present, and the muted labels sampled at `(165,170,180)` — my own eyeball had them wrong, the pixels settled it.

`REFERENCE.md` covers the magnitude-only rule and the junction-gap rule.

Full suite green: 3415 passed, 0 failed. No need to post them inline — this is enough to work from.

DIMOSUS:
Both fixed in 9ab57ef. Note the report is against `470b651`; the three points from the round before it were already fixed in `c83ea3a` — your own re-read of that head confirms two of them, and the third is the one below.

**1. Total loss survived the junction fix.** Correct, and it contradicted documentation I wrote in the same change, which is the worse half. `GetAdjacentPairs` declining to invent the Sub/Rear pair only removed the per-junction row; `BuildEntries` mints the total from the crossover window regardless, and the loss curve is drawn independently of either.

A view whose summed chain holds no handover now drops the loss entirely — curve and total. The **Sum stays**: the drivers do both play, and their combined output is a real thing to look at; it is the *loss* that claims a crossover exists.

I did not do this by making `LossChainZone` return null for `RearAndSub`. That would encode a fact about the reference car's crossover choices into the view model — a rear pair crossed at 110 to match its subs is a perfectly ordinary install, and it has a real junction. The question is about the loaded channels, so the code asks them: `ProcessedChannels.HasJunction`, a named predicate rather than an inline condition, because the curve and the read-out both have to obey it and a rule spelled twice will eventually disagree with itself.

**2. Stereo Δ ignored the selector.** Correct, and it is squarely the invariant this PR claims. It takes the shown blocks now.

**Left alone on purpose, so it is a decision and not an oversight:** the Sum loss toggle stays interactive in a view that *can* quote a loss but whose loaded channels happen not to cross. Whether a view quotes a loss is a property of the view; whether this project has a junction is a property of its data, and muting a control on data would make it flicker as sources resolve. Same reason the Sum toggle stays live with a single channel. Say the word if you would rather it followed the data.

`REFERENCE.md` covers both: the dropped loss with the Sum kept, and the read-out following the selector.

Full suite green: 3416 passed, 0 failed.

DIMOSUS:
Confirmed and fixed in a669da4 — and you are right that my test was too simple to show it. Two channels cannot express the case; three can, and the reference car IS three.

`HasJunction` answered the wrong question for the total. `Sub1 (<50) → Sub2 (50-110) → hole → Rear (>290)` holds a genuine junction, so the predicate said yes and the total came back for a set that is not one chain.

Took your second suggestion rather than the first, for the reason you gave: the two subwoofers' handover is real and worth reading, and losing it to fix the total would be paying information for correctness that can be had without it. So the two questions are now separate predicates:

- `HasJunction` still gates the **loss curve and the read-out** — with no handover anywhere, there is nothing to measure at all.
- `IsContinuousChain` gates the **total alone**, and requires every neighbour along the spectrum to hand over to the next (`pairs.Count == channels.Count - 1`, as you suggested).

On the reference Rear + Sub view that now gives: the `Sub1/Sub2` row shown, no `Sub2/Rear` row, no total, and the loss curve drawn — where it is honest at 50 Hz and simply flat where only one member plays.

The new test is the `[Sub1, Sub2, Rear]` set you asked for. It asserts both predicates disagree on it, that the surviving pair is specifically `Sub1/Sub2`, and that dropping the rear fill makes the same two subwoofers continuous again — so the rule cannot be satisfied by accident.

Full suite green: 3417 passed, 0 failed. `REFERENCE.md` now separates the two conditions as well.

PR #150, opened 2026-08-30
Settle the front chain first, then place the other groups against it

DIMOSUS:
Third of the "complex installations" series, on top of #147 (zones) and #148
(grouped views). Those two gave the tool the fact and the eyes; this one gives
Auto delay the staging.

## Why

The alignment engine walks ONE chain along the spectrum, pairing neighbours and
settling each junction against the last. That is the right shape for a crossover
chain and the wrong shape for a car. On the reference installation the rear pair
is high-passed at 290 Hz, so **by band centre it lands between the midrange and
the tweeter** — and the walk declared a midrange handing over to a rear fill at
the midrange's own low-pass corner. No filter does that. Auto delay was not
solving the problem badly; it was solving one nobody has.

## What

**Stage 1 — the front chain.** The front stage and the subwoofers under it,
walked by the existing engine, unchanged. The subwoofers belong here because
this is where their junctions are: on the reference car the two subs cross each
other and the lower one crosses the midbass.

**Stages 2 and 3 — placement, not search.** There is nothing between a rear fill
and the front stage to search: no filter hands a band from one to the other, and
a centre has no junction with anything. What those groups need is one number
each, and one number is what a correlation against a settled reference gives.
Two steps, the same shape the engine uses at a junction: a coarse band-limited
arrival difference says roughly where the answer is, then the phase-whitened
correlation is refined in a short window centred there. The coarse step is what
makes it work — a rear fill can sit twenty milliseconds from the front stage,
where a correlation searched around zero finds a lobe and one searched across
the whole range has many to choose between.

- A **rear fill** is read per side against its own side's front stage: left and
  right are different drivers at different distances, so forcing one delay on
  both would be the opposite of a tune. Then it is pushed back by the Rear fill
  offset.
- A **centre** has no side, so it is read against BOTH front sums and placed at
  the midpoint. The two readings are each other's witness — they should differ
  by the scene offset, since that is how far apart the sides are — and a
  disagreement is reported rather than averaged away.
- Polarity comes out of the same correlation as the sign of its extremum, so it
  is measured rather than guessed. It is applied only where no fill offset is in
  play: past a few milliseconds the two groups no longer sum in any way the ear
  resolves, and flipping there is cosmetics on a measurement that has stopped
  describing what is heard.
- A group the run cannot measure keeps its current delay, written EXPLICITLY
  into the override map. An absent override means zero to the reprocessor, so
  silence would have moved the very group the run just admitted it could not
  read.

**Stage 4 — normalization.** Every delay is relative until this pass, and freely
negative: a rear fill pushed back past the front asks the front to go negative.
The pass slides everything at once until the earliest sits at zero, preserving
every relation the stages computed. It does nothing to a set that is already
dialable, so no tune gains latency it did not ask for.

**The L/R bridge is now required to be a front-chain pair.** It is where the two
sides are tied together, and tied at a rear pair the whole scene would be
anchored to the fill instead of to the stage it is supposed to sit behind.

## Why the staging is safe rather than merely ordered

Only the front chain searches. A stage that searched would be free to move what
it was measured against; one that computes a placement from a settled reference
cannot disagree with it. So the stages are one-way by construction — no later
group can pull the front stage out of tune, other than by sliding all of it at
once in the normalization. That property is asserted directly in
`VirtualCrossoverAlignmentStageTests`.

## The dialog

A **Rear fill** row: how far behind the front stage the rear should arrive,
defaulting to the precedence-effect 15 ms, off entirely when the project has no
Rear block. Its tooltip carries the working numbers — the 10–20 ms range, that 0
co-arrives them for a second row of listeners, and the warning that rear
speakers often sit CLOSER to the ears than the front, so some delay is needed
just to reach zero and this offset is on top of that. The setting is stored with
the tune: a car tuned for its front seats and one tuned for its second row want
different answers.

**The rear's LEVEL is deliberately not here.** It sits 6–12 dB under the front
and the last word is the ear's; #148's `vs Front` block already reports the
current figure to adjust against, and teaching the gain engine about groups is a
change to a validated component that this PR does not need. Both the tooltip and
`REFERENCE.md` say so, so the absence is something a reader can act on rather
than discover.

## Merge bar

**The seven-cabin session battery is identical line for line** — the only diff
being the two wall-clock timings it prints. Run twice, once on this branch and
once on a stashed baseline, on both the single-side and the stereo path.

That is by construction rather than by luck: a project with no rear and no
centre reports that it needs no staging, and the caller then passes a null walk
set, which is literally the old engine call, and skips both later passes. Such a
project never enters the staged path at all.

The falsifier is `TheWalkNeverPairsAFrontDriverWithTheRearFill`, which records
the band ordering that caused the defect (`A, C, B` — the rear between the two
front drivers) and asserts the walk set now holds the front chain alone.

Full suite green: 3438 passed, 0 failed.

## Still ahead

The per-group crossover wizard (#4 of the plan) and the grouped tuning sheet
plus a per-device delay ceiling (#5). Field validation on the reference car is
the owner's, and its rear pair will need re-pointing from Front to Rear first —
the v9 migration cannot tell a rear pair from a front one.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -1,4 +1,4 @@
-# Professional Car Audio Tuning with Resonalyze
+﻿# Professional Car Audio Tuning with Resonalyze
 
 **A complete step-by-step guide**
 
@@ -649,8 +649,9 @@ The zone is what the **Show** selector under the plot sorts by, so on a front-on
 you can set every block to Front and Sub and forget it — that is the default view and
 what this chapter's four-way looks like throughout. On a larger install the zone is how
 you look at one part at a time: the front stage with its subwoofers, then the rear fill,
-then the centre. Auto delay does not read the zone yet, so on such a system treat its
-proposal as a starting point for the front stage only.
+then the centre. **Auto delay** reads the zone as well: it settles the front chain first
+and then places the rear fill and the centre against it, so on a larger install the zones
+are what make its proposal mean anything.
 
 So seven RAW measurements become four Virtual DSP channels:
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1694,10 +1694,8 @@ Each channel runs through:
   The zone decides what the main plot is about — see
   [Show](#show-which-part-of-the-installation-the-plot-is-about), which sorts the
   blocks by it and takes the curves, the Sum, the loss and the read-out with it.
-  Auto delay does not read it yet: it still walks every channel as one chain
-  along the spectrum, ordered by band centre, so on an installation with a rear
-  fill or a centre its proposal is not to be trusted until the staged alignment
-  lands. Projects saved
+  [Auto delay](#auto-delay) reads it too: the front chain is settled first and
+  the other groups are placed against it. Projects saved
   before zones existed are given one on load, guessed from what those files do
   record: a stereo pair becomes Front, and a mono block becomes Sub unless it
   high-passes, which no subwoofer does — that one becomes Center. Nothing else
@@ -2299,6 +2297,41 @@ actually achievable after the best per-junction delay.
 
 ### Auto delay
 
+#### Staging: the front chain first, then the groups against it
+
+An installation with a rear fill or a centre is not one crossover chain, so
+**Auto delay** does not walk it as one. It settles the **front chain** — the
+front stage and the subwoofers under it — with the search described below,
+unchanged. Then it places each further group against that result: one reading
+per group, and no re-tuning of anything already settled. Walked as one chain
+those groups produce junctions that do not exist, which on a car with a rear
+fill means a front midrange "handing over" to it at the midrange's own low-pass
+corner.
+
+The later groups are placed rather than searched, because there is nothing
+between them and the front stage to search: no filter hands a band from one to
+the other. A **rear fill** is read against the front stage of its own side (its
+left and right are different drivers at different distances, so each gets its
+own delay) and then pushed back by the **Rear fill** offset in the dialog. A
+**centre** has no side at all, so it is read against BOTH front sums and placed
+at the midpoint; the two readings are each other's witness, since they should
+differ by the scene offset, and a disagreement is reported instead of averaged
+away. Relative polarity comes out of the same correlation as its sign — but it
+is only applied where no fill offset is in play, because past a few
+milliseconds the two groups no longer sum in any way the ear resolves.
+
+A group the run cannot measure keeps the delay it had, and says so in the trace.
+
+Every delay is relative until the last pass, which slides the whole set until
+the earliest channel sits at zero — a rear fill pushed back past the front stage
+asks the front to go negative, and nothing is dialable until that shift. Every
+relation the stages computed survives it.
+
+A project with no rear and no centre never enters any of this: it takes the
+single-stage path, which is the same engine call it always took.
+
+#### The search
+
 **Auto delay** aligns in two stages: band-limited first arrivals, refined by a
 GCC-PHAT cross-correlation whose dominant extremum of either polarity seeds the
 junction (an inverted junction — a subwoofer against its midbass is the classic
@@ -2380,6 +2413,22 @@ sides of a pair by one shared delta to recover what the pin cost. Each far-side
 channel may also leave its own scene position by up to 0.03 ms to buy back its
 junction's summation — below what the scene can resolve, and enough to close the
 skew the arithmetic bridge leaves behind.
+
+**Rear fill** sets how far behind the front stage the rear arrives, and is off
+unless the project has a block in the Rear zone. Ten to twenty milliseconds
+(the default is fifteen) is where the precedence effect keeps the image on the
+dash while the rear adds room; zero co-arrives them, which is what a second row
+of listeners wants and what collapses the image for the front seats. Note that
+rear speakers are often CLOSER to your ears than the front, so some delay is
+needed just to reach zero and this offset sits on top of that. The setting is
+stored with the tune, since a car tuned for its front seats and one tuned for
+its second row want different answers.
+
+The rear's LEVEL is not set here. It usually sits 6-12 dB under the front —
+raise it until you notice it as a separate source, then take 2-3 dB back — and
+the last word belongs to the ear; the read-out's
+[vs Front block](#show-which-part-of-the-installation-the-plot-is-about) gives
+the current figure to adjust against.
 
 A run only proposes: the dialog answers with a report and nothing is written
 until **Apply**. It carries a row per channel — a value the run changes reads
--- a/source/Tools/VirtualCrossover/VirtualCrossoverAlignmentStage.cs
+++ b/source/Tools/VirtualCrossover/VirtualCrossoverAlignmentStage.cs
@@ -1,0 +1,103 @@
+namespace Resonalyze;
+
+/// <summary>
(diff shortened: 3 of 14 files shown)


DIMOSUS:
Self-review of my own branch before yours. Three defects found and fixed in b9b51a5; battery still identical.

**1. RHD mirror in the rear placement.** The group sums reach `PlaceLaterStagesStereo` in the engine's ROLES — on right-hand drive the reference is the right side — but the loop chose between them with `bool far = side.RightSide`, the cabin side. So every RHD run timed each rear driver against the *opposite* side's front stage. Now a named `IsFarSide(rightSide, rightHandDrive)` with all four cases pinned in a theory; this repository has been bitten by an unmirrored role before and a bare boolean is how it happens.

**2. The report never stated the rear fill offset.** It is the single most consequential number in a rear-fill tune, and a saved proposal could not say whether the rear had been co-arrived or held back. One line, now printed whenever it is non-zero.

**3. The one that mattered: a multi-driver later group was scattered.** A two-way rear or two-way centre had each of its drivers placed against the front stage *independently*. Each landed near its own right answer, and the junction BETWEEN them was overwritten by two unrelated numbers — worse than not touching them, since an untouched junction is at least still the one the user tuned. Such a group is now walked by the engine on its own junctions first, and the group placement is added on top, so a settled group moves as one. The plan for this series said a two-way centre "should just work"; it did not, and I had not checked.

**Two things I looked at and deliberately left:**

- `NormalizeStagedDelays` can in principle shift the set past the engine's own 50 ms ceiling. In practice it cannot get near it — the shift only occurs when a placement is negative, which needs the group arriving later than the front by more than its fill offset, and cabin path differences are a few ms. The per-device delay ceiling is PR 5's subject and the honest place for the check; the shift is printed in the trace meanwhile.
- The Sum-loss-style question of whether `passOctaves` in the placement means total width or half — checked, it is total (`halfOctaves = passOctaves / 2`), so `Log2(high/low)` gives exactly the intended band.

Full suite: 3443 passed, 0 failed. Battery identical line for line, single-side and stereo.

DIMOSUS:
Both reviews land on the same four, and Codex's inline set matches the agent's list. All four fixed — two in `b9b51a5` (my own self-review, which reached the same RHD and multi-block findings independently) and the remaining two plus the P2s in `4993bd0`.

**P1 — sparse-map normalization. Confirmed, and it is the worst defect in the branch.** The engine documents the convention I broke in as many words: *"A channel with no entry and a zero result keeps NO entry — absence means 'nothing proposed' (the reference)"*. Shifting only the map's keys left the anchor at zero while its siblings moved around it, so the "rigid-body" shift rewrote the front chain's internal alignment — the one relation in the run that must not change. It now takes the participant scope, reads absent as zero, and writes every participant back.

The pointed part of your note is the one I want to acknowledge: **my test could never have caught it, because it put the anchor in the dictionary by hand.** The replacement leaves it out, which is the state a real run is in, and asserts the front chain's gap survives the shift. A second test pins that an absent anchor also counts toward the minimum.

**P1 — RHD rear mirror. Confirmed**, and already fixed in `b9b51a5` before your review arrived; I found it reading the branch back. On the suggested alternative — passing physical `chainLeft`/`chainRight` instead of roles — I went the other way and kept the roles with a named `IsFarSide(rightSide, rightHandDrive)`. The reason is the centre: its witness compares the two readings against `sceneOffsetMs`, whose sign is defined in reference/far roles, so the roles have to survive into the placement anyway. Making the rear read them correctly is a smaller surface than making the centre read physical sides and re-derive the sign.

**P2 — band from the reference side only. Confirmed.** Each rear side now uses its own front chain's band. The centre uses the intersection of both, precisely because its two readings are averaged and compared: taken over different bands they would not be comparable.

**P2 — LOW CONFIDENCE invisible before Apply. Confirmed**, and the sharpest of the two P2s, since the failure mode is a doubtful centre being applied *because it looked ordinary*. Every placed channel now carries an `AlignmentDecision` — `Search` with a confidence for a placement, `Locked` with a reason for one the run refused — so the report shows it.

**Agent's #5, multi-block Rear/Center. Confirmed and already fixed** in `b9b51a5`. Both paths now walk a multi-driver group's own junctions through the engine first and add the group offset on top, so a two-way rear or centre keeps its internal alignment and moves as one. The two paths agree: single-side groups all members of a stage, stereo groups them by cabin side, which is the same rule applied to the sides that exist.

Session battery identical line for line after all of it. Full suite: 3445 passed, 0 failed.

On the suggested synthetic runs: sparse anchor with a negative placement, and the RHD mapping, are now unit-pinned. Asymmetric crossover bands and a two-way rear are covered structurally rather than end to end — the end-to-end path needs a panel, so those remain owner-side field checks on the reference car.

*(Apologies for the mangled duplicate above — shell backticks. This is the intended text.)*

DIMOSUS:
Confirmed and fixed in `7a9a7bf`. You are right that it is a certain crash, and it is worse than that as a mistake on my part: I wrote it into the *same commit* that fixed the identical convention violation in the normalization pass, one function away.

`AutoAlignmentEngine.Compute` leaves no entry for the channel it picks as the group's reference — absence means "nothing proposed" — so `inner[member]` throws `KeyNotFoundException` for exactly the case the multi-way fix had just introduced. Two drivers in a later group is the first time that code path can be reached at all, which is why nothing caught it.

Rather than change three call sites to `GetValueOrDefault`, I collapsed them into one named helper. The loop is a single line and it earns a name for what it must *not* be: writing the composition out three times is how two of the copies came to disagree with the convention in the first place. `SettleWithinGroup` now documents in its `<returns>` that the map it hands back is sparse.

**The regression test is the one you asked for.** It builds a real two-way rear — a low-pass and a high-pass crossing at 300 Hz, with the tweeter given a 0.6 ms head start — runs it through the real `AlignmentReprocessor` and the real engine, and then asserts three things:

- the map that comes back is genuinely sparse (`inner.Count < members.Count`), so the *reason* the helper exists is recorded rather than assumed — with the old indexer this is the assertion that proves the throw;
- the composition gives every member an override, the absent one reading as an untouched zero;
- the junction the engine just settled survives the copy, checked as the gap between the two drivers rather than their absolute values.

A second, engine-free test states the same rule directly, so it outlives any change in the engine's own behaviour.

Session battery identical line for line. Full suite: 3447 passed, 0 failed.

Worth saying plainly: two of the three defects in this branch have now been the same one — a sparse map read as if it were dense. The contract is documented in the engine and I broke it twice in a row. The helper and its `<returns>` note are there so the third time is harder.

DIMOSUS:
Fixed in `09d1f8b` rather than deferred — it is small, and it is the same mistake in miniature as the ones before it.

My own commit message said the line exists because *"a saved proposal could not say whether the rear had been co-arrived or held back"*. Guarded on `RearFillOffsetMs > 0` it could not say the first of those two — the exact case it was added to disambiguate. Zero is a choice, not an absence: it is what a second row of listeners wants, and a report silent about it reads identically to one where the rear was held back fifteen milliseconds.

The line is now printed whenever the run has a rear, read off the outcomes' zones as you suggested rather than off the offset, and it says `Rear fill 0.0 ms: co-arriving with the front stage.` in that case. A front-only project still says nothing, since the offset is stored per project and carries a value there too — three tests cover the three states.

Full suite: 3450 passed, 0 failed. Session battery identical line for line. Head is `09d1f8b`.

Thanks for the three rounds on this one — between them and my own pass, six defects came out of a branch whose CI was green throughout.

PR #153, opened 2026-08-31
Store IR sample arrays as base64 float32

DIMOSUS:
## What

The five bulk sample arrays of an impulse-response file (sweep-deconvolution real/imaginary, transfer real/imaginary, coherence) are now stored as a base64 string of **little-endian float32** values instead of indented JSON numbers. Everything else in the file stays readable text; in memory the properties remain `double[]`, so all analysis after a load runs in double exactly as before.

## Why

Written as indented JSON numbers those arrays cost ~28 bytes per sample and are the megabytes of the file. Measured on a real capture (Passat v2, 108.3 MB, re-saved via a scratchpad harness):

| | before | after |
|---|---|---|
| file size | 108.3 MB | **22.1 MB** (4.9x) |
| load | 757 ms | **27 ms** (28x) |
| save | - | 26 ms |
| max rounding error | - | -150...-169 dB |

Float32 keeps ~7 significant digits (-140 dB relative) - far below any measured noise floor. Every reloaded sample was verified to be exactly the float32 rounding of the original across all 4.3M values.

## Compatibility

- **Backward**: the converter reads both representations. A pre-v8 number array is read at its **full double precision** - the legacy path does not inherit the float32 rounding.
- **Forward**: format version bumped to 8. Unlike the additive metadata that deliberately never bumped it, this changes the representation of existing fields, so an older build must fail by version, not by parse error.
- Byte order is fixed little-endian by contract, not by host.

## Tests

1941 green. Added: the base64 block is byte-checked against independently encoded floats; a legacy v7 file with a value not representable in float32 loads it unrounded; a base64 block that is not a whole number of float32 values is rejected. Updated two pins: the round-trip coherence assert (0.91 is not float-exact) and the version pin in ArrayMicrophoneFileTests, with a note that v8 came from the sample representation, not from those sections. REFERENCE.md updated (format description + version).

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -954,14 +954,20 @@ its duration and sample count; the playback channel and measurement mode; the
 sweep-deconvolution samples plus the optional loopback transfer-function samples;
 optional coherence (γ²) data with the run counts; the stored meter values; an
 optional [SPL calibration anchor](#sound-pressure-level-db-spl); and embedded
-preview frequency-response data for the History panel.
+preview frequency-response data for the History panel. The bulk sample arrays —
+the megabytes of the file — are stored as base64-encoded little-endian float32
+rather than JSON numbers, which roughly quarters the file; float32 keeps about
+seven significant digits (−140 dB relative), far below any measured noise floor,
+and everything after a load still computes in double precision. Every other field
+stays readable text, and files written by earlier versions, with their plain
+number arrays, still load — at the full precision they were saved with.
 
 Click **Load** to open a previously saved response. Resonalyze validates the file
 first, rejects files below `44100 Hz`, restores the measurement metadata into the
 active record, and redraws the current view from the loaded data — without
 rewriting the audio-device configuration. Saving and loading are disabled while a
 measurement is running. The current file format identifier is
-`resonalyze-impulse-response`, version `7`.
+`resonalyze-impulse-response`, version `8`.
 
 ### Importing a sweep recorded elsewhere
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -967,7 +967,10 @@ first, rejects files below `44100 Hz`, restores the measurement metadata into th
 active record, and redraws the current view from the loaded data — without
 rewriting the audio-device configuration. Saving and loading are disabled while a
 measurement is running. The current file format identifier is
-`resonalyze-impulse-response`, version `8`.
+`resonalyze-impulse-response`, version `8`. A file declaring a LATER version than
+the build knows is refused by that version, up front — a newer version exists to
+change how something is stored, and reading on would report a broken file where
+"made by a newer Resonalyze" is the answer.
 
 ### Importing a sweep recorded elsewhere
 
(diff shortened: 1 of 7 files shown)


DIMOSUS:
Review finding addressed in c6856d1.

The finding was right, and half of it was an overclaim in my own comment: builds up to v7 validate the version only after full deserialization and cannot be changed now, so on a v8 file they do fail with a parse error — no fix can reach them. What CAN be fixed is this build forward: `LoadAsync` now preflights the declared `format` + `version` with the streaming `JsonFormatMarker` probe (extended to read the version marker alongside the format, same chunked walk, no deserialization) and refuses a later version up front with "Unsupported impulse response version N". From v8 on, a future format change reads as a version refusal, never as a broken file.

Pinned by tests: a version-9 file whose samples use a representation v8 does not know (an object) is rejected with "version 9", not a `JsonException`; plus probe tests for version-before-format ordering, nested versions not counting, and missing/non-numeric versions. The comment at `CurrentVersion` and REFERENCE.md now state the honest behavior. 1947 app tests green.

DIMOSUS:
Re the P1 "Keep analyzed IR samples in double precision" — we measured it instead of arguing. The finding conflates two different things, and the experiment below shows the distinction empirically.

**What AGENTS.md's rule protects** is intermediate computation past the analysis boundary — dividing tiny spectral values, subtracting near-equal phases, accumulating millions of terms. This PR does not touch any of that: the properties stay `double[]`, and every FFT, group delay, GCC-PHAT and summation after a load runs in double exactly as before. What the PR quantizes is the stored **input** data — a one-time relative error of ~1e-7 (−150 dB), on measurements whose capture buffers are float32 to begin with (AGENTS.md: "Raw and real-time audio samples stay float") and whose acoustic noise floor sits tens of dB above that.

**The experiment.** The session battery is the most cancellation-sensitive consumer in the repository: it loads 7 archived cabins (76 impulse-response files, 2993 MB), computes the Auto delay proposal on each session's own chains (GCC-PHAT, junction combing, coherence, group delay), and judges both the proposal and the owner's saved tuning by the panel's own summation-loss metric, printing a fixed-format report meant to be diffed line by line. We ran it twice on identical code: once on the original double-precision files, once on a copy where every IR was re-saved through the v8 float32 format (absolute paths inside the copied sessions rewritten so the locator could not silently fall back to the originals).

**Result.** Out of 184 report lines:

- all 92 proposal/detail lines — every proposed delay (0.01 ms), every front, every direct-PHAT lobe correlation and lag, every polarity — **identical**;
- 25 of 27 high-precision ROW lines (metrics printed to six decimal places of a dB) — **identical**;
- 2 ROW lines differ in one column by **0.000001 dB**: the v3 B/C junction dip reads −2.319213 instead of −2.319214, and the cabin's total row inherits that same digit.

So the full float32 round trip moved the battery's most precise printed figure by one unit in the sixth decimal place of a dB, once across seven cabins, and changed no decision, no proposal and no readout at panel precision. The panel displays 0.01 dB / 0.01 ms; the observed effect is five orders of magnitude below that.

Keeping the finding unapplied. The float32 encoding stays; the analysis boundary it worries about was never crossed.

PR #155, opened 2026-08-31
Reclassify below-noise harmonics: neutral note instead of the amber overlap warning

DIMOSUS:
## Problem

In FR mode, a harmonic that fell **below the measurement noise floor** triggered the threatening amber warning "HDn packet overlaps its neighbour — increase the sweep duration", scolding users for their **cleanest** captures. The overlap check reads window edges relative to the packet's own peak, so it cannot tell a leaking packet from no packet at all: a noise-only window has edges ~10–12 dB below its plateau maximum (sigma·sqrt(2·ln M) crest factor), squarely inside the "overlap" verdict.

## Fix

**DSP** (`EssHarmonicAnalysis`)
- A robust time-domain tail-noise amplitude (same quiet region `EssNoise` draws from; median of 8 chunk RMS values, so a stray thump cannot inflate it).
- A packet whose plateau peak is within 16 dB of that floor (covers the noise crest factor up to M ~ 10^7 with margin) holds no harmonic: new `IsBelowNoiseFloor` classification — still dropped from curves/THD (its "curve" would be the noise floor), but **no warning**, and `IsValid` stays true.
- No usable tail (short or silent record) → the below-noise test stands down and the old verdict applies unchanged. Genuine leaks sit far above the floor and keep the amber warning.

**UI** (`PlotModelFactory`, `MeasurementPlotContext`)
- Missing HD curves split into two buckets: real problems keep the amber annotation; below-noise orders get a neutral gray note — *"below the measurement noise floor / distortion too low to resolve — a clean capture, not a fault"*. Mixed cases stack both, each naming only its own curves.

**Rendering** (`OverlayTextAnnotation`)
- Long-standing bug: multi-line overlay text was vertically centred on its anchor line, pushing the first line half out of the plot area (visible on the old amber warning too). The block is now anchored by its flowing edge — single-line notes (legend labels, SPL/transfer notices, live overlays) render pixel-identical, multi-line blocks grow into the plot and can no longer clip.
- The long overlap-warning line wraps at "; " so narrow windows fit it.

## Tests

- 5 new DSP tests (`EssHarmonicBelowNoiseTests`): below-noise classification with no warnings, drop-without-warning through the curve pipeline, genuine overlap NOT reclassified when noise is present, a harmonic well above the floor untouched, fallback to the old verdict without a usable tail.
- 2 new UI tests (`PlotModelFactoryTests`): gray-note-only for a clean capture; mixed case renders amber (HD2) + gray (HD3, HD4) with correct colors and bucketing.
- Full suites green: DSP 1415, App 1955.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -240,7 +240,7 @@ neighbouring order, so the curve is dropped from the plot and from THD and an
 amber warning names the order and the cure: a longer sweep, or a narrower
 analysed range (marginal isolation keeps the curve and just carries the
 caveat). If instead the whole window holds **nothing above the record's own
-noise floor** — plateau and edges alike — there is no harmonic to draw at all:
+noise floor** — anywhere in it — there is no harmonic to draw at all:
 the curve is dropped the same way, but the note is a neutral gray *"below the
 measurement noise floor"*, because distortion too low for the capture to
 resolve is the mark of a clean measurement, not a fault. A longer sweep or more
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -231,6 +231,23 @@ frequency (a second-harmonic hump from a 1 kHz drive appears at 1 kHz, not at
 sweep-deconvolution scale while the primary curve is loopback-normalized, their
 vertical distance is not yet a calibrated distortion percentage.
 
+A ticked HD curve can still be missing from the plot, and the plot says why at
+its top centre rather than leaving the checkbox unexplained. Each order's
+harmonic lives in its own time window of the deconvolved sweep, and two things
+can empty that window of a drawable curve. If the order's packet has **not
+decayed by its window edge** it leaks into — or is polluted by — the
+neighbouring order, so the curve is dropped from the plot and from THD and an
+amber warning names the order and the cure: a longer sweep, or a narrower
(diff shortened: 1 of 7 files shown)


DIMOSUS:
Both P1s addressed in 374441d:

- **Edge contamination**: the below-noise verdict now also requires both edge RMS values to sit within 6 dB of the tail-noise floor. A neighbour's tail in an edge region over a noise-level plateau falls through to the warned overlap verdict instead of being blessed as clean. Pinned by a new contaminated-edge test.
- **Docs**: REFERENCE.md (Frequency Response / distortion curves section) now documents both ways a ticked HD curve can vanish — the amber overlap warning and the gray below-noise note, including that the two can appear together.

DIMOSUS:
Shoulder gap closed in the latest commit: the below-noise verdict now takes the maximum over the WHOLE window (the 16 dB margin already budgets the noise crest factor at that length), instead of only the plateau and the two edge regions. The 6 dB edge-RMS bound stays as a second, tighter check for low-crest coherent contamination the maximum could miss. A new test pins a burst in the shoulder between the plateau and the trailing edge, with both formerly-checked regions at the noise floor; REFERENCE.md wording updated to match ("anywhere in it").

